Issue #21, opened 2020-10-05
 [FEATURE] Add output option to detect_breakends & find_deletions


Irallia:
There are already tests that can test this: https://github.com/seqan/iGenVar/pull/19

PR #37, opened 2020-11-02
[FEATURE, TEST] Making find_deletions write to an output file and not to std::out. And adding tests.

simonsasse:
Resolves 1/2 of https://github.com/seqan/iGenVar/issues/21

### DIFF
--- a/include/find_deletions/deletion_finding_and_printing.hpp
+++ b/include/find_deletions/deletion_finding_and_printing.hpp
@@ -2,7 +2,6 @@
 
 #include <seqan3/std/filesystem>
 #include <fstream>
-#include <iostream>
 #include <vector>
 
 #include "junction.hpp"
@@ -10,7 +9,8 @@
 /*! \brief Detects deletions out of the junction file.
  *
  * \param junction_file_path input junction file
+ * \param output_file_path output vcf file
  *
  * \details Extracts deletions out of given breakends / junctions.
  */
-void find_and_print_deletions(const std::filesystem::path & junction_file_path);
+void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
--- a/src/find_deletions.cpp
+++ b/src/find_deletions.cpp
@@ -5,6 +5,7 @@
 struct cmd_arguments
 {
     std::filesystem::path junction_file_path{};
+    std::filesystem::path output_file_path{};
 };
 
 void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments & args)
@@ -12,7 +13,10 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments 
     parser.info.author = "David Heller";
     parser.info.short_description = "Find deletions";
     parser.info.version = "0.0.1";
-    parser.add_positional_option(args.junction_file_path, "Input junctions tab-separated format.");
+    parser.add_option(args.junction_file_path, 'i', "input", "Input junctions tab-separated format.",
+                      seqan3::option_spec::REQUIRED);
+    parser.add_option(args.output_file_path, 'o', "output", "The path of the vcf output file.",
+                      seqan3::option_spec::REQUIRED );
 }
 
 int main(int argc, char ** argv)
@@ -29,7 +33,7 @@ int main(int argc, char ** argv)
         seqan3::debug_stream << "[Error] " << ext.what() << "\n";       // customise your error message
         return -1;
     }
-    find_and_print_deletions(args.junction_file_path);
+    find_and_print_deletions(args.junction_file_path, args.output_file_path);
 
 
     return 0;
--- a/src/find_deletions/deletion_finding_and_printing.cpp
+++ b/src/find_deletions/deletion_finding_and_printing.cpp
@@ -1,12 +1,13 @@
 #include "find_deletions/deletion_finding_and_printing.hpp"
 
+#include <fstream>
 /*! \brief Reads the input junction file and stores the junctions in a vector.
  *
  * \param junction_file_path input junction file
  *
  * \returns a vector of junctions
  */
-std::vector<junction> read_junctions(const std::filesystem::path & junction_file_path)
+std::vector<junction> read_junctions(std::filesystem::path const & junction_file_path)
 {
     std::fstream junction_file;
     junction_file.open(junction_file_path, std::ios::in);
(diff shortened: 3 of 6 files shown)


test/api/deletion_finding_and_printing_test.cpp
```diff
@@ -1,18 +1,26 @@
...
-    std::string expected{"##fileformat=VCFv4.2\n"
-                         "##source=iGenVarCaller\n"
-                         "CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO\n"
-                         "chr21	9435236	.	N	<DEL>	60	PASS	SVTYPE=DEL;SVLEN=-50;END=9435286\n"
-                         "chr21	11104574	.	N	<DEL>	60	PASS	SVTYPE=DEL;SVLEN=-248;END=11104822\n"};
+    find_and_print_deletions(DATADIR"detect_breakends_shorted.vcf", DATADIR"find_deletions_file_out_test.vcf");
+    std::ifstream f;
+    f.open(DATADIR"find_deletions_file_out_test.vcf");
+    std::stringstream buffer;
+    buffer << f.rdbuf();
```
Irallia, line 15: I got the tip from jörg to use EXPECT_RANGE_EQ when comparing a test file. With you you don't need to create a test file for the three lines, but you can probably use that here and it will be shorter?
My code from `test/cli/detect_breakends_cli_test.cpp`:
```cpp
TEST_F(cli_test, test_outfile)
{
    cli_test_result result = execute_app("detect_breakends",
                                         data("simulated.minimap2.hg19.coordsorted_cutoff.sam"),
                                         "detect_breakends_insertion_file_out.fasta");

    std::filesystem::path out_file_path = "detect_breakends_insertion_file_out.fasta";
    std::filesystem::path test_file_path = "../../data/detect_breakends_insertion_file_test.fasta";
    seqan3::sequence_file_input out_file{out_file_path};
    seqan3::sequence_file_input test_file{test_file_path};

    EXPECT_RANGE_EQ(out_file, test_file);
}
```
my idea:
```suggestion
    find_and_print_deletions(DATADIR"detect_breakends_shorted.vcf", DATADIR"find_deletions_file_out_test.vcf");

    std::filesystem::path out_file_path = "find_deletions_file_out_test.vcf";
    seqan3::sequence_file_input out_file{out_file_path};
```
But I don't know whether that is possible and whether "expected" has to be packed differently ...

test/api/deletion_finding_and_printing_test.cpp
```diff
@@ -1,18 +1,26 @@
...
-    EXPECT_EQ(expected, std_cout);
+    std::string expected
+    {
+	    "##fileformat=VCFv4.2\n"
+            "##source=iGenVarCaller\n"
+            "CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO\n"
+            "chr21\t9435236\t.\tN\t<DEL>\t60\tPASS\tSVTYPE=DEL;SVLEN=-50;END=9435286\n"
+    };
+    EXPECT_TRUE(f.is_open());
+    EXPECT_EQ(buffer.str(), expected);
```
Irallia, line 25: And then
```suggestion
    EXPECT_RANGE_EQ(out_file, expected);
```

src/find_deletions.cpp
```diff
@@ -13,6 +14,7 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
     parser.info.short_description = "Find deletions";
     parser.info.version = "0.0.1";
     parser.add_positional_option(args.junction_file_path, "Input junctions tab-separated format.");
+    parser.add_positional_option(args.output_file_path, "The path of the output file.");
```
Irallia, line 17: I think we should use argument flags here:
```suggestion
    parser.add_option(args.junction_file_path, 'i', "input", "Input junctions tab-separated format.", seqan3::option_spec::REQUIRED);
    parser.add_option(args.output_file_path, 'o', "output", "The path of the vcf output file.", seqan3::option_spec::REQUIRED);
```
@eldariont: Or do you think that this is overhead because this is a call of a part of iGenVar and we will connect these program parts?

Irallia (changes requested):
Thanks for your work!
Could you please rebase on the current master?
Most of my requests are more like discussion points.

src/find_deletions/deletion_finding_and_printing.cpp
```diff
@@ -70,16 +76,18 @@ void find_and_print_deletions(const std::filesystem::path & junction_file_path)
                         int32_t distance = mate2_pos - mate1_pos;
                         if (distance > 40 && distance < 100000)
                         {
-                            print_deletion(junctions[i].get_mate1().seq_name, mate1_pos, mate2_pos, 60);
-                        }
+                            print_deletion(junctions[i].get_mate1().seq_name, mate1_pos, mate2_pos, 60, out_file);
+                            out_file.close();
```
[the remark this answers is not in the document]
simonsasse: Good point. Guess I didn't really notice that there is a for loop under all those ifs ! ;)

src/find_deletions.cpp
simonsasse: I added them :)

test/api/deletion_finding_and_printing_test.cpp
simonsasse: So I found out that the expect_range_eq only makes sense if we use the vcf parser of SeqAn because expect_range_eq compares ranges over records and if we just read the file with a stream, we don't have any records. 
So should we add the vcf parser or leave it as is ?

Irallia: You are absolutely right. Leave it as it is and we will do that when the vcf parser is there.
It still has to be implemented in seqan ;)

Irallia (approved):
Wonderful, thank you very much!

test/CMakeLists.txt
```diff
@@ -35,13 +35,13 @@ macro (add_app_test test_filename test_alternative)
     # Create the test target.
     add_executable (${target} ${test_filename})
     target_link_libraries (${target} "${PROJECT_NAME}_lib" seqan3::seqan3 gtest_all pthread)
+    target_include_directories ("${target}" PUBLIC ${CMAKE_SOURCE_DIR}/lib/seqan3/test/include)
```
smehringer, line 38: I would put this change into a separate `[INFRA]` commit. This way you can later on find all you changes to the infra structure more easily, instead of them being mixed with other normal commits.

smehringer (changes requested):
May I suggest something that does not need to be changed or part of this PR:
You go from functions `foo()` that print to std::cout to function `foo(std::ofstream & out_file)` that prints to a file. If you would do 
```cpp
template<typename stream_type>
void foo(stream_type & stream)
{
    stream << // ...
}
```
You can pass either a file OR std::cout. Writing to an dedicated output file is always very helpful in most application but I find sometimes, when writing pipelines, it can be very handy if the file can also be written to std::cout. This of course depends on your App and whether you think it may or may not be executed as part of a pipeline. If not you can ignore my suggestion, if so you might want to consider writing to `cout` unless a output_path is given. :) Just ideas 

Note that there is also a output_stream seqan3 concept for checking that is has the `<<` operator.

include/find_deletions/deletion_finding_and_printing.hpp
```diff
@@ -10,7 +10,8 @@
 /*! \brief Detects deletions out of the junction file.
  *
  * \param junction_file_path input junction file
+ * \param output_file_path output vcf file
  *
  * \details Extracts deletions out of given breakends / junctions.
  */
-void find_and_print_deletions(const std::filesystem::path & junction_file_path);
+void find_and_print_deletions(const std::filesystem::path & junction_file_path, const std::filesystem::path & output_file_path);
```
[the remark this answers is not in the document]
Irallia: Since I don't have any strong feelings, I would prefer the seqan3 style.

smehringer (approved):

PR #39, opened 2021-01-18
Add overload function for output to std::cout. Remove mandatory requi…

### LINKED ISSUE #38: Extend the output option to decide if we want to write to a file and or to std::cout.
This issue must be discussed: 

The question is, do we want to be able to choose whether we want to write an output file and or be able to write in std::cout aswell. The following review comment gives a general example and further information.

From a comment in https://github.com/seqan/iGenVar/pull/37#pullrequestreview-530999883
> May I suggest something that does not need to be changed or part of this PR:
> You go from functions `foo()` that print to std::cout to function `foo(std::ofstream & out_file)` that prints to a file. If you would do
> 
> ```c++
> template<typename stream_type>
> void foo(stream_type & stream)
> {
>     stream << // ...
> }
> ```
> 
> You can pass either a file OR std::cout. Writing to an dedicated output file is always very helpful in most application but I find sometimes, when writing pipelines, it can be very handy if the file can also be written to std::cout. This of course depends on your App and whether you think it may or may not be executed as part of a pipeline. If not you can ignore my suggestion, if so you might want to consider writing to `cout` unless a output_path is given. :) Just ideas
> 
> Note that there is also a output_stream seqan3 concept for checking that is has the `<<` operator.

joshuak94:
…rement for output file option. Add testing.

Resolves #38. Output file is no longer a mandatory option. If called without an output file given, will output to std::cout by default.

Note: I didn't do this very efficiently... Created an overload function which resulted in a bit of code duplication, but I couldn't think of a clever way to allow passing a filesystem path OR std::cout without opening the ostream in main.cpp.

### DIFF
--- a/include/find_deletions/deletion_finding_and_printing.hpp
+++ b/include/find_deletions/deletion_finding_and_printing.hpp
@@ -1,7 +1,9 @@
 #pragma once
 
 #include <seqan3/std/filesystem>
+#include <seqan3/io/stream/concept.hpp>
 #include <fstream>
+#include <iostream>
 #include <vector>
 
 #include "junction.hpp"
@@ -14,3 +16,6 @@
  * \details Extracts deletions out of given breakends / junctions.
  */
 void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
+
+//!\overload
+void find_and_print_deletions(std::filesystem::path const & junction_file_path);
--- a/include/find_deletions/deletion_finding_and_printing.hpp
+++ b/include/find_deletions/deletion_finding_and_printing.hpp
@@ -18,4 +18,5 @@
 void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::ostream & out_stream);
 
 //!\overload
-void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
+void find_and_print_deletions(std::filesystem::path const & junction_file_path,
+                              std::filesystem::path const & output_file_path);
--- a/include/find_deletions/deletion_finding_and_printing.hpp
+++ b/include/find_deletions/deletion_finding_and_printing.hpp
@@ -11,11 +11,11 @@
 /*! \brief Detects deletions out of the junction file.
  *
  * \param junction_file_path input junction file
- * \param output_file_path output vcf file
+ * \param out_stream output stream
  *
  * \details Extracts deletions out of given breakends / junctions.
  */
-void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
+void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::ostream & out_stream);
 
 //!\overload
-void find_and_print_deletions(std::filesystem::path const & junction_file_path);
+void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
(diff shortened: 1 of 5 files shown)


joergi-w (changes requested):
I make a suggestion to avoid the code duplication.

src/find_deletions/deletion_finding_and_printing.cpp
```diff
@@ -121,3 +131,42 @@ void find_and_print_deletions(std::filesystem::path const & junction_file_path,
     }
     out_file.close();
 }
+
+//!\overload
+void find_and_print_deletions(std::filesystem::path const & junction_file_path)
```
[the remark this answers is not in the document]
joshuak94: !! Thanks for this! I knew there was a better way, I just couldn't think of a simple way to do it...

src/find_deletions/deletion_finding_and_printing.cpp
```diff
@@ -41,13 +41,13 @@ std::vector<junction> read_junctions(std::filesystem::path const & junction_file
 
 /*! \brief Prints the header of a vcf file to a given outputfile.
  *
- * \param out_file ouput file ofstream object
- */   
-void print_vcf_header(std::ofstream & out_file)
+ * \param out_stream ouput stream object
```
joergi-w, line 44:
```suggestion
 * \param out_stream output stream object
```

joergi-w (approved):
Looks great! 
I found one typo:

include/find_deletions/deletion_finding_and_printing.hpp
```diff
@@ -1,16 +1,21 @@
...
  * \param junction_file_path input junction file
- * \param output_file_path output vcf file
+ * \param out_stream output stream
  *
  * \details Extracts deletions out of given breakends / junctions.
  */
+void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::ostream & out_stream);
+
+//!\overload
 void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path);
```
Irallia, line 21: Can you insert a line break? We want to have a limit of 120 characters per line.

src/find_deletions.cpp
```diff
@@ -15,8 +15,7 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
     parser.info.version = "0.0.1";
     parser.add_option(args.junction_file_path, 'i', "input", "Input junctions tab-separated format.",
                       seqan3::option_spec::REQUIRED);
-    parser.add_option(args.output_file_path, 'o', "output", "The path of the vcf output file.",
-                      seqan3::option_spec::REQUIRED );
+    parser.add_option(args.output_file_path, 'o', "output", "The path of the vcf output file. If no path is given, will output to standard output.");
```
Irallia, line 18: line breack please 💅

src/find_deletions/deletion_finding_and_printing.cpp
```diff
@@ -56,68 +56,82 @@ void print_vcf_header(std::ofstream & out_file)
...
+                         }
+                     }
+                 }
+             }
+         }
+     }
+ }
+
+//!\overload
 void find_and_print_deletions(std::filesystem::path const & junction_file_path, std::filesystem::path const & output_file_path)
```
Irallia, line 121: line breack plz 💅 I know some of them were already there. 😅

Irallia (approved):
Thank you! 
Could you add to your PR description: `Resolves #38` ?
And I found some style things, nothing important, but maybe yo have some time for new lines. ❤️

Irallia (approved):
Thank you!

PR #45, opened 2021-01-20
[FEATURE] Modularization of the methods for detecting junctions.

### LINKED ISSUE #40: [FEATURE] Modularize methods for detecting junctions

Irallia:
Resolves #40 
Adds also some new tests for different method combinations.

### DIFF
--- a/include/detect_breakends/junction_detection.hpp
+++ b/include/detect_breakends/junction_detection.hpp
@@ -12,6 +12,7 @@
  *
  * \param alignment_file_path input file - path to the sam/bam file
  * \param insertion_file_path output file - path for the fasta file
+ * \param methods list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
  *          We sort out unmapped alignments, secondary alignments, duplicates and alignments with low mapping quality.
@@ -19,4 +20,5 @@
  *          For primary alignments, also the split read information is analyzed.
  */
 void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_file_path,
-                                        const std::filesystem::path & insertion_file_path);
+                                        const std::filesystem::path & insertion_file_path,
+                                        const std::vector<uint8_t> methods);
--- a/include/detect_breakends/junction_detection.hpp
+++ b/include/detect_breakends/junction_detection.hpp
@@ -12,7 +12,10 @@
  *
  * \param alignment_file_path input file - path to the sam/bam file
  * \param insertion_file_path output file - path for the fasta file
- * \param methods list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
+ * \param methods - list of methods for detecting junctions (1: cigar_string,
+ *                                                           2: split_read,
+ *                                                           3: read_pairs,
+ *                                                           4: read_depth)
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
  *          We sort out unmapped alignments, secondary alignments, duplicates and alignments with low mapping quality.
--- a/src/detect_breakends.cpp
+++ b/src/detect_breakends.cpp
@@ -6,17 +6,38 @@ struct cmd_arguments
 {
     std::filesystem::path alignment_file_path{};
     std::filesystem::path insertion_file_path{};
+    std::vector<uint8_t> methods{1, 2, 3, 4};   // default is using all methods
 };
 
 void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments & args)
 {
-    parser.info.author = "David Heller";
+    parser.info.author = "David Heller & Lydia Buntrock";
+    parser.info.app_name = "iGenVar";
+    parser.info.man_page_title = "Short and long Read SV Caller";
     parser.info.short_description = "Detect junctions in a read alignment file";
     parser.info.version = "0.0.1";
+    parser.info.date = "19-01-2021";    // last update
+    parser.info.email = "lydia.buntrock@fu-berlin.de";
+    parser.info.long_copyright = "long_copyright";
+    parser.info.short_copyright = "short_copyright";
+    parser.info.url = "https://github.com/seqan/iGenVar/";
+
+    // Validatiors:
+    // seqan3::value_list_validator method_validator{"1", "cigar_string",
+    //                                               "2", "split_read",
+    //                                               "3", "read_pairs",
+    //                                               "4", "read_depth"};
+    seqan3::arithmetic_range_validator method_validator{1, 4};
+
+    // Options - Input / Output:
     parser.add_positional_option(args.alignment_file_path, "Input read alignments in SAM or BAM format.",
                                  seqan3::input_file_validator{{"sam", "bam"}} );
     parser.add_positional_option(args.insertion_file_path, "Output file for insertion alleles",
                                  seqan3::output_file_validator{{"fa", "fasta"}} );
+
+    // Options - Methods:
+    parser.add_option(args.methods, 'm', "method", "Choose the method to be used.",
+                      seqan3::option_spec::ADVANCED, method_validator);
 }
 
 int main(int argc, char ** argv)
@@ -24,16 +45,21 @@ int main(int argc, char ** argv)
     seqan3::argument_parser myparser{"detectJunctions", argc, argv};    // initialise myparser
     cmd_arguments args{};
     initialize_argument_parser(myparser, args);
+
+    // Parse the given arguments and catch possible errors.
     try
     {
         myparser.parse();                                               // trigger command line parsing
     }
     catch (seqan3::argument_parser_error const & ext)                   // catch user errors
     {
-        seqan3::debug_stream << "[Error] " << ext.what() << "\n";       // customise your error message
+        seqan3::debug_stream << "[Error] " << ext.what() << '\n';       // customise your error message
         return -1;
     }
-    detect_junctions_in_alignment_file(args.alignment_file_path, args.insertion_file_path);
+
+    detect_junctions_in_alignment_file(args.alignment_file_path,
+                                       args.insertion_file_path,
+                                       args.methods);
 
     return 0;
 }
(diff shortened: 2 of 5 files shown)


src/detect_breakends.cpp
```diff
@@ -6,34 +6,60 @@ struct cmd_arguments
...
+    parser.info.email = "lydia.buntrock@fu-berlin.de";
+    parser.info.long_copyright = "long_copyright";
+    parser.info.short_copyright = "short_copyright";
+    parser.info.url = "https://github.com/seqan/iGenVar/";
+
+    // Validatiors:
+    // seqan3::value_list_validator method_validator{"1", "cigar_string",
+    //                                               "2", "split_read",
+    //                                               "3", "read_pairs",
+    //                                               "4", "read_depth"};
```
Irallia, line 29: I was thinking, if the user wants to insert names instead of numbers, so we could offer both.

test/api/junction_detection_test.cpp
```diff
@@ -18,8 +18,87 @@ TEST(junction_detection, fasta_out_not_empty)
 {
     std::string expected{
         "Reference\tchr22\t17458417\tForward\tReference\tchr21\t41972615\tForward\tm41327/11677/CCS\n"
+        "Reference\tchr22\t17458417\tForward\tReference\tchr21\t41972615\tForward\tm41327/11677/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm21263/13017/CCS\n"
         "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm21263/13017/CCS\n"
         "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm38637/7161/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm38637/7161/CCS\n"
+        "Reference\tm2257/8161/CCS\t41972616\tForward\tRead \t0\t2294\tForward\tchr21\n"
+        "Reference\tm2257/8161/CCS\t41972616\tReverse\tRead \t0\t3975\tReverse\tchr21\n"
```
Irallia, line 27: I don't understand why this changed and we have some duplications now.
@eldariont maybe you have an idea?

src/detect_breakends.cpp
```diff
@@ -6,34 +6,60 @@ struct cmd_arguments
...
+
+    // Options - Input / Output:
     parser.add_positional_option(args.alignment_file_path, "Input read alignments in SAM or BAM format.",
                                  seqan3::input_file_validator{{"sam", "bam"}} );
     parser.add_positional_option(args.insertion_file_path, "Output file for insertion alleles",
                                  seqan3::output_file_validator{{"fa", "fasta"}} );
+
+    // Options - Methods:
+    parser.add_option(args.methods, 'm', "method", "Choose the method to be used.",
+                      seqan3::option_spec::ADVANCED, method_validator);
```
joshuak94, line 40: The way this is, a user could only choose one of the four methods, no? E.g. you can't use `-m 1, 2, 3, 4` or something. Maybe instead, it's better to have four flags instead of options? Then the user could do something like `-c -s -d` for cigar, split reads, and read depth.

Although to be completely honest, I can't really think of a reason why a user would not want to use the CIGAR string. For split reads, if the user has very short reads I guess they might think split reads aren't useful. Same for read depth if the sequencing was very shallow.

With regards to read pairs, I think this is something we should detect automatically (whether the sequencing was done with single end or paired end sequencing), so that a user doesn't accidentally turn on the read pairs option with single end sequencing data.

joshuak94 (changes requested):
Looks good! We should figure out if it really makes sense to have all four of the options. Also, duplicate entries in the test cases definitely need to be solved...

include/detect_breakends/junction_detection.hpp
```diff
@@ -12,11 +12,13 @@
  *
  * \param alignment_file_path input file - path to the sam/bam file
  * \param insertion_file_path output file - path for the fasta file
+ * \param methods list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
```
eldariont, line 15:
```suggestion
 * \param methods - list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
```

include/detect_breakends/junction_detection.hpp
```diff
@@ -12,11 +12,13 @@
...
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
  *          We sort out unmapped alignments, secondary alignments, duplicates and alignments with low mapping quality.
  *          Then, the CIGAR string of all remaining alignments is analyzed.
  *          For primary alignments, also the split read information is analyzed.
  */
 void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_file_path,
-                                        const std::filesystem::path & insertion_file_path);
+                                        const std::filesystem::path & insertion_file_path,
+                                        const std::vector<uint8_t> methods);
```
eldariont, line 24: If I understand correctly, this function implements all four detection methods but can be parametrized by the `methods` parameter to only use a subset of methods. This makes a lot of sense because the alignment file has to be processed only once (in the best case) and different methods can be applied on the same read in one go.

However, this architecture is not really modularized, right? To be really modularized we would have four `detect_junctions`functions that each implement one method. But then, the alignment file would be processed four times in the worst case. So I guess that's why you decided to choose the other option?!

test/api/junction_detection_test.cpp
eldariont: Yes, I think it's because your switch clause is missing the `break` statements so that for `method=1` both case 1 and case2 (and cases 3 and 4) are executed.

src/detect_breakends.cpp
eldariont: I think with the current implementation, a user can choose four methods using something like `-m 1 -m 2 -m 3 -m 4` because args.methods is a `std::vector`.

You have a good point, though, with the CIGAR strings and the read pairs. It might be good to spend some time thinking about efficient processing of the SAM/BAM files:

- cigar and split reads collect evidence from each individual read.
- read depth, in contrast, is an analysis performed on the depth profile along the genomes, not individual reads.
- read pair should (like you say) only be used for paired-end data. But for this type of data it works on a per-read basis like cigar and split reads.

So we should maybe have a separate detection function for read depth because it's completely separate from the other methods.

src/detect_breakends/junction_detection.cpp
```diff
@@ -220,6 +220,7 @@ void analyze_cigar(std::string chromosome,
  * \cond
  * \param alignment_file_path input file - path to the sam/bam file
  * \param insertion_file_path output file - path for the fasta file
+ * \param methods list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
```
eldariont, line 223:
```suggestion
 * \param methods - list of methods for detecting junctions (1: cigar_string, 2: split_read, 3: read_pairs, 4: read_depth)
```

test/api/junction_detection_test.cpp
```diff
@@ -18,8 +18,87 @@ TEST(junction_detection, fasta_out_not_empty)
...
+}
+
+TEST(junction_detection, method_1_only)
+{
+    std::string expected{
+        "Reference\tchr22\t17458417\tForward\tReference\tchr21\t41972615\tForward\tm41327/11677/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm21263/13017/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm38637/7161/CCS\n"
+        "Reference\tm2257/8161/CCS\t41972616\tForward\tRead \t0\t2294\tForward\tchr21\n"
+        "Reference\tm2257/8161/CCS\t41972616\tReverse\tRead \t0\t3975\tReverse\tchr21\n"
```
eldariont, line 52: If only method 1 is used I would expect only calls from the CIGAR strings (deletions and insertions). The first three lines here are translocation breakends which are probably output because of the missing `break` statements.

test/api/junction_detection_test.cpp
```diff
@@ -18,8 +18,87 @@ TEST(junction_detection, fasta_out_not_empty)
...
+
+TEST(junction_detection, method_1_and_2)
+{
+    std::string expected{
+        "Reference\tchr22\t17458417\tForward\tReference\tchr21\t41972615\tForward\tm41327/11677/CCS\n"
+        "Reference\tchr22\t17458417\tForward\tReference\tchr21\t41972615\tForward\tm41327/11677/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm21263/13017/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm21263/13017/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm38637/7161/CCS\n"
+        "Reference\tchr22\t17458418\tForward\tReference\tchr21\t41972616\tForward\tm38637/7161/CCS\n"
```
eldariont, line 101: The duplicates should be removed (see above).

src/detect_breakends.cpp
Irallia: I think my change (the modularisation) is there to compare all methods with all methods. 
But later, such a recognition from the SAM/BAM files would certainly be useful. I would create a new issue for this.
And then you could make some methods dependent on the input files.

test/api/junction_detection_test.cpp
Irallia: Stupid me 😅

include/detect_breakends/junction_detection.hpp
Irallia: Yes, that is correct. Maybe the term modularize is not really the right one.
The idea is, that we can combine different methods. And compare them. As we also want to compare run time, the complete decoupling would end in a higher run time if we combine two methods.
So I would leave it and would change the term modularisation with decoupling?

joshuak94 (approved):
Looks good! I'm still more of a fan of using flags instead of numbered methods (I think `-c -s -d` looks more descriptive/intuitive than `-m 1 -m 2 -m 4`), but it's largely a stylistic choice so both are fine!

include/detect_breakends/junction_detection.hpp
eldariont: I see. But I think it's not really a decoupling either. How about "Enable selection of methods for detecting junctions" as new title for this pull request?

eldariont (approved):
Looks all good now :)

Issue #29, opened 2020-08-07
iGenVar - Add / Use a VCF format parser from SeqAn3


Irallia:
Add a vcf format parser to such that writing vcf can be done more easily.

Irallia:
We wrote our own VCF parser which will some day moved to the seqan3 library -> https://github.com/seqan/product_backlog/issues/302

PR #46, opened 2021-01-22
Add a skeleton VCF parser which can print variants in VCF output.

### LINKED ISSUE #9: [FEATURE] SeqAn3: Develop a simple VCF parser.
VCF: Variant Call Format, a file format standard in bioinformatics.
https://en.wikipedia.org/wiki/Variant_Call_Format

Simple means a parser in which the content of the columns is not checked (e.g. without specific info fields).
We will do this in a follow up issue.

### LINKED ISSUE #10: [FEATURE] SeqAn3: Develop info fields of the VCF parser.
VCF: Variant Call Format, a file format standard in bioinformatics.

Any key in the info field is allowed, although some subfields are reserved (albeit optional), see wikipedia:
https://en.wikipedia.org/wiki/Variant_Call_Format

We want to use an enum for the reserved subfields.

joshuak94:
Resolves #9 and resolves #10.

Here I introduce a skeleton VCF parser to address issue #29. It doesn't really do any checking to make sure values are valid, but it does provide a template for information regarding variants to be stored in a VCF format-friendly way. 

Some things up for discussion:
1. Currently, the INFO header is stored in a vector of info_entry objects. In the variants themselves, the respective INFO entries are stored in a map, with the key being the INFO name and the value being the INFO value... I'm not too satisfied with this current implementation (to have a std::map for every variant seems unnecessary).
2. Constructors for `variant_header` and `variant_record` are currently empty, because I figured all of the information is not always known on construction. But maybe it makes sense to change this?
3. The INFO field can store various types of information (integers, floats, flags, characters, and strings). Currently, everything is being stored as a string as I wasn't sure the best way to do this.

### DIFF
--- a/include/variant_parser/variant_record.hpp
+++ b/include/variant_parser/variant_record.hpp
@@ -34,7 +34,7 @@ public:
 
     /*! \brief Set the file format for this VCF file header.
      *
-     *  \param fileformati The input fileformat.
+     *  \param fileformat_i The input fileformat.
      */
     void set_fileformat(std::string fileformat_i)
     {
@@ -43,12 +43,12 @@ public:
 
     /*! \brief Add header information for a given INFO field.
      *
-     *  \param info_keyi The INFO key name.
-     *  \param numberi The number of values this key can hold.
-     *  \param typei The type of values this key holds.
-     *  \param descriptioni The description of this INFO field..
-     *  \param sourcei The source of the INFO field.
-     *  \param versioni The version of the source.
+     *  \param info_key_i The INFO key name.
+     *  \param number_i The number of values this key can hold.
+     *  \param type_i The type of values this key holds.
+     *  \param description_i The description of this INFO field..
+     *  \param source_i The source of the INFO field.
+     *  \param version_i The version of the source.
      */
     void add_meta_info(std::string info_key_i, std::uint8_t number_i, std::string type_i, std::string description_i,
                        std::string source_i, std::string version_i)
@@ -97,7 +97,7 @@ public:
 
     /*! \brief Set the chromosome for a variant.
      *
-     * \param chromi The chromosome value to use.
+     * \param chrom_i The chromosome value to use.
      */
     void set_chrom(std::string chrom_i)
     {
@@ -106,7 +106,7 @@ public:
 
     /*! \brief Set the pos for a variant.
      *
-     * \param posi The pos value to use.
+     * \param pos_i The pos value to use.
      */
     void set_pos(std::uint64_t pos_i)
     {
@@ -115,7 +115,7 @@ public:
 
     /*! \brief Set the id for a variant.
      *
-     * \param idi The id value to use.
+     * \param id_i The id value to use.
      */
     void set_id(std::string id_i)
     {
@@ -124,7 +124,7 @@ public:
 
     /*! \brief Set the ref for a variant.
      *
-     * \param refi The ref value to use.
+     * \param ref_i The ref value to use.
      */
     void set_ref(std::string ref_i)
     {
@@ -133,7 +133,7 @@ public:
 
     /*! \brief Set the alt for a variant.
      *
-     * \param alti The alt value to use.
+     * \param alt_i The alt value to use.
      */
     void set_alt(std::string alt_i)
     {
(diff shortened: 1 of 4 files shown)


eaasna (commented):
Nive work!

include/variant_parser/variant_record.hpp
```diff
@@ -0,0 +1,193 @@
+#include <seqan3/io/stream/concept.hpp>
```
eaasna, line 1: What's the connection to the seqan3 VCF parser? Will you also be using it?

src/find_deletions/deletion_finding_and_printing.cpp
```diff
@@ -97,22 +72,38 @@ void find_and_print_deletions(std::filesystem::path const & junction_file_path,
...
                     int32_t mate1_pos = junctions[i].get_mate1().position;
                     int32_t mate2_pos = junctions[i].get_mate2().position;
+                    int32_t length{};
                     if (junctions[i].get_mate1().orientation == strand::forward)
                     {
                         int32_t distance = mate2_pos - mate1_pos;
                         if (distance > 40 && distance < 100000)
                         {
-                            print_deletion(junctions[i].get_mate1().seq_name, mate1_pos, mate2_pos, 60, out_file);
+                            length = mate2_pos - mate1_pos;
```
eaasna, line 88: Why do deletions have to be dealt with separately? (compared to SNV)

eaasna (commented):
Thanks!

include/variant_parser/variant_record.hpp
joshuak94: Hi! So at the moment, seqan3 doesn't have a VCF parser implemented. So the one I'm implementing here will be used as a sort of framework for the future seqan3 implementation!

src/find_deletions/deletion_finding_and_printing.cpp
Irallia: At the moment, we're just managing deletions. The plan is to have an example (deletions) to go though the whole workflow.
The other SNVs will be managed later.

include/variant_parser/variant_record.hpp
```diff
@@ -0,0 +1,193 @@
...
+    variant_header() {}
+
+    /*! \brief Set the file format for this VCF file header.
+     *
+     *  \param fileformati The input fileformat.
+     */
+    void set_fileformat(std::string fileformati)
+    {
+        fileformat = fileformati;
+    }
```
Irallia, line 31: Just an idea, so it doesn't look like a typo:
```suggestion
    /*! \brief Set the file format for this VCF file header.
     *
     *  \param fileformat_i The input fileformat.
     */
    void set_fileformat(std::string fileformat_i)
    {
        fileformat = fileformat_i;
    }
```

include/variant_parser/variant_record.hpp
```diff
@@ -0,0 +1,193 @@
+#include <seqan3/io/stream/concept.hpp>
+#include <fstream>
+#include <map>
+
```
Irallia, line 4: I would add a short description like
```cpp
/*
 * We're using the The Variant Call Format (VCF) Specification: VCFv4.3 and BCFv2.2
 * https://samtools.github.io/hts-specs/VCFv4.3.pdf
 */
 ```

File: include/variant_parser/variant_record.hpp [hunk 5]
```diff
@@ -0,0 +1,193 @@
...
+class info_entry
```
Irallia, line 5: And maybe something like this for the info_enty class:
```cpp
/*
 * An Info field looks like:
 * ##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version">
 * https://samtools.github.io/hts-specs/VCFv4.3.pdf (1.4.2 Information field format)
 */
```

include/variant_parser/variant_record.hpp
eaasna: I see, thanks for the explanation!

src/find_deletions/deletion_finding_and_printing.cpp
eaasna: Ah, I see. Thanks for explaining :)

eaasna (approved):
Looks good!

Irallia (changes requested):
Thanks for your work!
Could you maybe add some Information to the classes and the head of the file, as I mentioned? 
And if you don't have a strong opinion, I would replace that `i` with an `_i`. 💅  
Thanks again!

Irallia (approved):
LGFM, thank you!

PR #50, opened 2021-02-04
[FEATURE] modularisation refining breakends

### LINKED ISSUE #42: [FEATURE] Modularize methods for refining breakends

Irallia:
Resolves #42 .

I added the new parser option -r with 2 different refining methods and the possibility to not use one. (none is implemented)

### DIFF
--- a/include/detect_breakends/junction_detection.hpp
+++ b/include/detect_breakends/junction_detection.hpp
@@ -23,6 +23,9 @@
  *                                                           1: hierarchical_clustering,
  *                                                           2: self-balancing_binary_tree,
  *                                                           3: candidate_selection_based_on_voting)
+ * \param refinement_method method for refining breakends (0: no_refinement,
+ *                                                         1: sViper_refinement_method,
+ *                                                         2: sVirl_refinement_method)
  * \param min_var_length - minimum length of variants to detect (default 30 bp)
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
@@ -34,4 +37,5 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
                                         const std::filesystem::path & insertion_file_path,
                                         const std::vector<uint8_t> methods,
                                         const clustering_methods clustering_method,
+                                        const refinement_methods refinement_method,
                                         const uint64_t min_var_length);
--- a/include/method_enums.hpp
+++ b/include/method_enums.hpp
@@ -8,3 +8,11 @@ enum clustering_methods
     self_balancing_binary_tree = 2,
     candidate_selection_based_on_voting = 3
 };
+
+//!\brief An enum for the different refinement methods.
+enum refinement_methods
+{
+    no_refinement = 0,
+    sViper_refinement_method = 1,
+    sVirl_refinement_method = 2
+};
--- a/src/detect_breakends.cpp
+++ b/src/detect_breakends.cpp
@@ -23,7 +23,24 @@ auto enumeration_names(clustering_methods)
                                                   {"3", clustering_methods::candidate_selection_based_on_voting},
                                                   {"candidate_selection_based_on_voting",
                                                    clustering_methods::candidate_selection_based_on_voting}};
-}
+};
+
+// Specialise a mapping from an identifying string to the respective value of your type refinement_methods. With the
+// help of this function, you're able to call ./detect_breackends with -r 0 and -r no_refinement and get the same
+// result.
+auto enumeration_names(refinement_methods)
+{
+    return std::unordered_map<std::string_view,
+                              refinement_methods>{{"0", refinement_methods::no_refinement},
+                                                  {"no_refinement",
+                                                   refinement_methods::no_refinement},
+                                                  {"1", refinement_methods::sViper_refinement_method},
+                                                  {"sViper_refinement_method",
+                                                   refinement_methods::sViper_refinement_method},
+                                                  {"2", refinement_methods::sVirl_refinement_method},
+                                                  {"sVirl_refinement_method",
+                                                   refinement_methods::sVirl_refinement_method}};
+};
 
 struct cmd_arguments
 {
@@ -31,6 +48,7 @@ struct cmd_arguments
     std::filesystem::path insertion_file_path{};
     std::vector<uint8_t> methods{1, 2, 3, 4};                   // default is using all methods
     clustering_methods clustering_method{simple_clustering};    // default is the simple clustering method
+    refinement_methods refinement_method{no_refinement};        // default is using no refinement
     uint64_t min_var_length = 30;
 };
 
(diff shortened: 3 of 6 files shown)


src/detect_breakends/junction_detection.cpp
```diff
@@ -365,6 +369,20 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
             break;
     }
 
+    switch (refinement_method)
+    {
+        case 0: // no refinement
+            seqan3::debug_stream << "No refinement was selected.\n";
+            break;
+        case 1: // sViper
+            seqan3::debug_stream << "The sViper refinement method is not jet implemented\n";
```
eaasna, line 378:
```suggestion
            seqan3::debug_stream << "The sViper refinement method is not yet implemented\n";
```

src/detect_breakends/junction_detection.cpp
```diff
@@ -365,6 +369,20 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
...
+            break;
+        case 2: // prioritzation_based_on_rank_aggregation
+            seqan3::debug_stream << "The prioritzation_based_on_rank_aggregation refinement method is not jet"
```
eaasna, line 381:
```suggestion
            seqan3::debug_stream << "The prioritzation_based_on_rank_aggregation refinement method is not yet"
```

include/detect_breakends/junction_detection.hpp
```diff
@@ -21,6 +21,9 @@
  *                                                           1: hierarchical_clustering,
  *                                                           2: self-balancing_binary_tree,
  *                                                           3: candidate_selection_based_on_voting)
+ * \param refinement_method method for refining breakends (0: no_refinement,
+ *                                                         1: sViper,
+ *                                                         2: prioritzation_based_on_rank_aggregation)
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
```
eaasna, line 28: I think that junctions and breakends represent places where the query sequence differs from the reference. But I'm not quite sure about the exact meaning of the two terms

eaasna (commented):
Nice work :) 

just a couple of typos (and a question below)

src/detect_breakends/junction_detection.cpp
Irallia: and I did it again ^^ thanks for the hint 😅

include/detect_breakends/junction_detection.hpp
eaasna: Thanks for the explanation! :)

eaasna (approved):

include/detect_breakends/junction_detection.hpp
```diff
@@ -22,6 +22,9 @@
  *                                                           1: hierarchical_clustering,
  *                                                           2: self-balancing_binary_tree,
  *                                                           3: candidate_selection_based_on_voting)
+ * \param refinement_method method for refining breakends (0: no_refinement,
+ *                                                         1: sViper,
+ *                                                         2: prioritzation_based_on_rank_aggregation)
```
joshuak94, line 27: Prioritization spelling

joshuak94: I'm not quite sure what you mean with the prioritization based on rank aggregation part. If this adapted from Vaquita, then it doesn't belong in refining breakpoints.

The prioritization used in Vaquita is more to "rank" the variants by how important they are. This also has the effect that the variants which are considered "more important" are more likely to be true variants. However, rank aggregation does not do anything to the existing breakpoints. It's sort of a "final step" in the pipeline.

For now, I would think that the only breakend refinement we have is via sViper.

src/detect_breakends.cpp
```diff
@@ -23,14 +23,32 @@ auto enumeration_names(clustering_methods)
...
+                                                   refinement_methods::prioritzation_based_on_rank_aggregation}};
+};
 
 struct cmd_arguments
 {
     std::filesystem::path alignment_file_path{};
     std::filesystem::path insertion_file_path{};
     std::vector<uint8_t> methods{1, 2, 3, 4};                   // default is using all methods
     clustering_methods clustering_method{simple_clustering};    // default is the simple clustering method
+    refinement_methods refinement_method{no_refinement};        // default is using no refinement
```
joshuak94, line 51: Why not make the default to using sViper?

joshuak94: Also we have to consider that sViper will only work if the user provides long reads AND short reads. With only one dataset given, the command parser should probably throw an error if sViper refinement is selected.

joshuak94 (changes requested):
Looks good! But to be honest, I don't know why we want to give this as an option. I think it would be fine to have this be automatically handled behind the scenes. AKA if only short or only long reads are given, then sViper is not run. However, if both are given, sViper should be run automatically.

Also the rank aggregation doesn't belong with breakpoint refinement, I think. It's just calculating a score basically based on what evident is present and how strongly it's present. It doesn't actually modify any of the variants.

include/detect_breakends/junction_detection.hpp
Irallia: Yes you are right, the `prioritzation_based_on_rank_aggregation` is probably wrong here. I spoke to david and we have still 2 different refining methods. The other is a method from SVirl.
On the other side, I think we want to able and enable the refinement, as it also depends on the input. sViper only works if we have short and long reads. The method from sVirl works with long reads, and we have none for short reads only.

We could remove this parameter later and could do this decision automaticaly but for now it is easyer for us and we are able to add more if there are more methods.

src/detect_breakends.cpp
Irallia: Yes, we have to solve https://github.com/seqan/igenvar/issues/43 next, and there we could add a check if sViper is possible to use when we know which input we have.

include/method_enums.hpp
```diff
@@ -8,3 +8,11 @@ enum clustering_methods
     self_balancing_binary_tree = 2,
     candidate_selection_based_on_voting = 3
 };
+
+//!\brief An enum for the different refinement methods.
+enum refinement_methods
+{
+    no_refinement = 0,
+    sViper = 1,
+    sVirl_refinement_method = 2
```
joshuak94, line 17: Small thing, we should say `sViper_refinement_method` to keep it consistent with the naming.

src/detect_breakends/junction_detection.cpp
```diff
@@ -373,6 +377,20 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
...
+    switch (refinement_method)
+    {
+        case 0: // no refinement
+            seqan3::debug_stream << "No refinement was selected.\n";
+            break;
+        case 1: // sViper
+            seqan3::debug_stream << "The sViper refinement method is not yet implemented\n";
+            break;
+        case 2: // prioritzation_based_on_rank_aggregation
+            seqan3::debug_stream << "The prioritzation_based_on_rank_aggregation refinement method is not yet"
```
joshuak94, line 389:
```suggestion
            seqan3::debug_stream << "The sVirl refinement method is not yet"
```

joshuak94 (changes requested):
I think there should also be tests to cover the sViper and sVirl refinement cases, even if they aren't yet implemented. Just so that in the future, any changes will be covered by the tests already.

Irallia:
You are right about the missing tests. I will add in an extra PR the others for the other parameters. I hope these are the tests you meant.

test/api/junction_detection_test.cpp
```diff
@@ -120,6 +125,65 @@ TEST(junction_detection, method_1_and_2)
...
+    testing::internal::CaptureStdout();
+    detect_junctions_in_alignment_file(DATADIR"simulated.minimap2.hg19.coordsorted_cutoff.sam",
+                                       tmp_dir/"detect_breakends_out_short.fasta",
+                                       {1, 2, 3, 4},
+                                       simple_clustering,
+                                       sViper_refinement_method,
+                                       sv_default_length);
+
+    std::string std_cout = testing::internal::GetCapturedStdout();
+    seqan3::debug_stream << "std_out:\n" << std_cout << '\n';
```
joshuak94, line 162: Is there a reason you decided to print the `std::out` output, but not actually test that it gives what's expected? Just wondering. I would say we should either test the output, or just remove the printing.

joshuak94 (approved):

test/api/junction_detection_test.cpp
Irallia: Doesn't the line after that test just that?

Issue #48, opened 2021-02-04
Tests don't compile on MacOS 11.1 (Big Sur)


eldariont:
Hi,

on a brand new installation of MacOS, I installed the current versions of Xcode, Command Line Tools, cmake (homebrew) and gcc@7 (homebrew). When I follow the installation steps in the README, `make` succeeds but `make test` fails with:

```
Test project /Users/eldarion/Documents/Projects/mpi/iGenVar-build
[ 27%] Built target datasource--simulated.minimap2.hg19.coordsorted_cutoff.sam
[ 36%] Built target iGenVar_lib
[ 60%] Built target googletest
[ 63%] Linking CXX executable junction_detection_test
ld: warning: object file (../googletest/src/googletest-build/lib/libgtest.a(gtest-all.cc.o)) was built for newer macOS version (11.0) than being linked (10.16.2)
ld: warning: object file (../googletest/src/googletest-build/lib/libgtest_main.a(gtest_main.cc.o)) was built for newer macOS version (11.0) than being linked (10.16.2)
Undefined symbols for architecture x86_64:
  "testing::internal::PrintStringTo(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::basic_ostream<char, std::char_traits<char> >*)", referenced from:
      testing::AssertionResult testing::internal::CmpHelperEQ<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >(char const*, char const*, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) in junction_detection_test.cpp.o
  "testing::internal::GetCapturedStdout[abi:cxx11]()", referenced from:
      junction_detection_fasta_out_not_empty_Test::TestBody()      in junction_detection_test.cpp.o
...
ld: symbol(s) not found for architecture x86_64
collect2: error: ld returned 1 exit status
make[4]: *** [test/api/junction_detection_test] Error 1
make[3]: *** [test/api/CMakeFiles/junction_detection_test.dir/all] Error 2
make[2]: *** [test/CMakeFiles/api_test.dir/rule] Error 2
make[1]: *** [test/CMakeFiles/api_test.dir/rule] Error 2
...
```

[...]

PR #61, opened 2021-02-11
[FEATURE] New class cluster for groups of similar junctions

### LINKED ISSUE #51: [FEATURE] Decouple junctions and a cluster of junctions
After we added a first simple clustering method
eldariont left a comment:
> However, I think that the implementation of the simple clustering method (including the corresponding changes to the junction class) could be improved. Proviously, the junction class represented a single junction detected from a single read. Now, it is also used to represent clusters of junctions. IMO, we should represent clusters of junctions with a decouple junctions and cluster of junctions. or a std::vector. Then, we could keep all the cluster-related information, such as the number of supporting reads, the read names, etc. separated from the information on a single junction.

We would like to create this new class.

eldariont:
This PR adds a new class `cluster` that represents a cluster of junctions by holding an `std::vector<junction>`. The new class implements public methods `get_average_mate1()` and `get_average_mate2` to obtain the average of all junctions in the cluster. The PR also updates the simple clustering method to use the new `cluster` class. The PR resolves #51.

I'm not very experienced with C++ and the coding guidelines yet. So please don't be shy in the reviews and point out all the potentially ugly parts of my code 🙂 Also: I was not able to run the tests on my machine (see #48) so they might be broken.

### DIFF
--- a/include/cluster.hpp
+++ b/include/cluster.hpp
@@ -1,6 +1,7 @@
 
 #include <utility>
 #include <cmath>
+#include <stdexcept>
 
 class cluster
 {
@@ -12,10 +13,10 @@ public:
      * \{
      */
     constexpr cluster()                    = default; //!< Defaulted.
-    cluster(cluster const &)              = default; //!< Defaulted.
-    cluster(cluster &&)                   = default; //!< Defaulted.
-    cluster & operator=(cluster const &)  = default; //!< Defaulted.
-    cluster & operator=(cluster &&)       = default; //!< Defaulted.
+    cluster(cluster const &)               = default; //!< Defaulted.
+    cluster(cluster &&)                    = default; //!< Defaulted.
+    cluster & operator=(cluster const &)   = default; //!< Defaulted.
+    cluster & operator=(cluster &&)        = default; //!< Defaulted.
     ~cluster()                             = default; //!< Defaulted.
     //!\}
 
@@ -23,19 +24,19 @@ public:
     {
         
     }
-    uint64_t get_cluster_size() const
+    size_t get_cluster_size() const
     {
         return members.size();
     }
 
     breakend get_average_mate1() const
     {
-        sequence_type seq_type;
-        std::string seq_name;
+        sequence_type seq_type{};
+        std::string seq_name{};
         uint64_t sum_positions = 0;
-        strand orientation;
+        strand orientation{};
         // Iterate through members of the cluster
-        for (uint32_t i=0; i<members.size(); i++)
+        for (uint32_t i = 0; i < members.size(); ++i)
         {
             breakend mate1 = members[i].get_mate1();
             if (i == 0)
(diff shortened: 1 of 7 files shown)


Irallia:
I would pass the first review to someone else (seqan/team) and do the second review. But one point I saw directly. It is common to start structs and classes with a capital letter. Can you adjust that?

joergi-w (changes requested):
The `cluster` class is not a class template, so I recommend moving the implementations into a cpp file and link them together. This can speed up compilation time a lot for your application.

eldariont:
Thanks @joergi-w for your great suggestions. I incorporated all of them except the one below:

> The `cluster` class is not a class template, so I recommend moving the implementations into a cpp file and link them together. This can speed up compilation time a lot for your application.

I see the point of doing this but don't know the perfect way to do it. Currently, we have the executable `detect_breakends.cpp` that requires the library `detect_breakends/junction_detection.cpp` which in turn uses `cluster.hpp`. I separated `cluster.hpp` into `cluster.hpp` (declarations) and `cluster.cpp` (definitions). If I now link the library `cluster.cpp` to the executable `detect_breakends.cpp`, I get the error that `detect_breakends/junction_detection.cpp` does not know the cluster class (i.e. one library does not see the another library). What am I making wrong? Sorry for this noob question.

joergi-w (approved):
Thank you, looks good!
The suggestion with the cpp file is not necessary for you code to work, it was more an idea that you can still incorporate at a later time. 

To detect the cluster class, make sure that "cluster.hpp" is included and the respective library has in cmake the right `target_include_path` to find it. Then all the definitions of the header should be known. If the linking breaks (shown as: Symbol not found error), then add the cpp files that contain the missing function definitions to the respective target.

include/cluster.hpp
```diff
@@ -0,0 +1,108 @@
+
+#include <utility>
+#include <cmath>
+#include <stdexcept>
```
Irallia, line 4: could you sort the includes lexicographically

include/cluster.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+class cluster
```
Irallia, line 6:
```suggestion
class Cluster
```
as we discussed, or we do this change in an extra PR where we change junctions and breakends aswell

Irallia (changes requested):
LGFM, just one tiny thing 💅

include/cluster.hpp
eldariont: Yes, I planned to create a separate PR to keep things tidy :)

Irallia (approved):

Irallia:
As a last comment: I would do this moving / splitting in cpp and hpp files in another PR (I already started with these things), as we should do this for the other classes etc aswell.

eldariont:
> As a last comment: I would do this moving / splitting in cpp and hpp files in another PR (I already started with these things), as we should do this for the other classes etc aswell.

Yes, I agree. I got the splitting of cluster working now locally but am not sure whether it's the optimal way to do it. Maybe it would be best if you start a PR and I have a look.

Irallia:
Ah and rebase and amend the last commit to have a tidy history before you merge ❤️

PR #63, opened 2021-02-15
[MISC] make iGenVar classes and structs start with uppercase letters

eldariont:
breakend -> Breakend
junction -> Junction
cluster -> Cluster
aligned_segment -> AlignedSegment


Irallia (approved):
LGFM, I think you can merge directly, we don't need a second review for these changes.

Issue #59, opened 2021-02-11
Method selection over CLI does not work properly


eldariont:
Hi,

for me, the selection of specific junction detection methods (probably the same for clustering methods) via the CLI does not work properly. As introduced with https://github.com/seqan/iGenVar/pull/45, the set of methods to use can be specified via the `-m` option. But when I add `seqan3::debug_stream << args.methods << '\n';` after the argument parsing:

```
> build/bin/detect_breakends simulated.minimap2.hg19.coordsorted_cutoff.sam insertion_file_out.fasta -m1
[1,2,3,4,1]
...
The read pair method is not yet implemented.
The read depth method is not yet implemented.
...
```

Apparently, the parser just appends values to the already initialized `std::vector<uint8_t> methods{1, 2, 3, 4};` and executes all methods even though `-m1` was called.

Best
David

Irallia:
I have already noticed the problem and have a solution lying around.

PR #64, opened 2021-02-15
[FEATURE] Switch methods to an enum vector and write own validator

Irallia:
Resolves https://github.com/seqan/iGenVar/issues/59
and changes the method input parameter to an enum with an associated custom validator.

Was blocked by https://github.com/seqan/seqan3/pull/2381
Was blocked by https://github.com/seqan/seqan3/pull/2394


src/detect_breakends.cpp
```diff
@@ -23,13 +37,13 @@ auto enumeration_names(clustering_methods)
...
                                                    clustering_methods::candidate_selection_based_on_voting}};
-}
+};
 
 struct cmd_arguments
 {
     std::filesystem::path alignment_file_path{};
     std::filesystem::path insertion_file_path{};
-    std::vector<uint8_t> methods{1, 2, 3, 4};                   // default is using all methods
+    std::vector<detecting_methods> methods{};
```
eseiler, line 46:
```suggestion
    std::vector<detecting_methods> methods{cigar_string, split_read, read_pairs, read_depth};  // default is using all methods
```

Or does that look funny in the help page of the argument parser?

include/method_enums.hpp
```diff
@@ -1,5 +1,14 @@
 #pragma once
 
+//!\brief An enum for the different clustering methods.
+enum detecting_methods
+{
+    cigar_string = 1,
+    split_read = 2,
+    read_pairs = 3,
+    read_depth = 4
+};
```
eseiler, line 10:
```suggestion
    read_depth = 4
    // Also add new methods to the default values in the argument parsers
};
```

Just as a reminder

src/detect_breakends.cpp
```diff
@@ -48,12 +62,12 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
...
 
     // Validatiors:
-    // seqan3::value_list_validator method_validator{"1", "cigar_string",
-    //                                               "2", "split_read",
-    //                                               "3", "read_pairs",
-    //                                               "4", "read_depth"};
-    seqan3::arithmetic_range_validator method_validator{1, 4};
-    seqan3::value_list_validator clustering_method_validator {(seqan3::enumeration_names<clustering_methods> | seqan3::views::get<1>)};
+    seqan3::value_list_validator detecting_method_validator {
+        (seqan3::enumeration_names<detecting_methods> | seqan3::views::get<1>)
```
eseiler, line 66:
```suggestion
        (seqan3::enumeration_names<detecting_methods> | std::views::values)
```

src/detect_breakends.cpp
```diff
@@ -48,12 +62,12 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
...
+    };
+    seqan3::value_list_validator clustering_method_validator {
+        (seqan3::enumeration_names<clustering_methods> | seqan3::views::get<1>)
```
eseiler, line 69:
```suggestion
        (seqan3::enumeration_names<clustering_methods> | std::views::values)
```

src/detect_breakends.cpp
```diff
@@ -91,6 +105,12 @@ int main(int argc, char ** argv)
         return -1;
     }
 
+    // default settings if no parameters given
+    if (args.methods.size() == 0)
+    {
+        args.methods = {cigar_string, split_read, read_pairs, read_depth};  // default is using all methods
+    }
+
```
eseiler, line 113: See above. I would set it as default in `args` unless it does some funny help page stuff.
If it does (and hence you want the default in `args` to be empty), the `add_option` should note that default = all methods.

include/method_enums.hpp
```diff
@@ -1,5 +1,14 @@
 #pragma once
 
+//!\brief An enum for the different clustering methods.
+enum detecting_methods
+{
+    cigar_string = 1,
```
eseiler, line 6: Why not start with 0?

include/method_enums.hpp
eseiler, line 10: ## If there may be more detecting methods in the future, you could think about doing this:

```suggestion
//!\brief An enum for the different clustering methods.
enum detecting_methods
{
    cigar_string = 0,
    split_read = 1,
    read_pairs = 2,
    read_depth = 3,
    //!\cond
    // ATTENTION: Must always be the last item; will be used to determine the number of ids.
    SIZE //!< Determines the size of the enum.
    //!\endcond
};
```

Then you know the size via `detecting_methods::SIZE`,
there could for example be a test that tests that  `args.methods.size() == detecting_methods::SIZE`, i.e. the default method selection uses all methods.

eseiler (changes requested):

src/detect_breakends.cpp
Irallia: No, I chose this way because otherwise we can never choose only one or two methods (see issue description). Before, the set methods were added to the default vector and not replaced the vector. If you know a better way, I'm all ears.

eseiler: Ah, that's very unintuitive and should be fixed if possible. https://github.com/seqan/seqan3/issues/2393

include/method_enums.hpp
Irallia: We use 0 for "do not use any method" (see refinement), so I decided to start with 1 here.

include/method_enums.hpp
Irallia: @eldariont does this sound usefull to you?

include/method_enums.hpp
```diff
@@ -1,5 +1,16 @@
...
```
eseiler, line 6:
```suggestion
{
    none = 0,
    cigar_string = 1,
```

include/method_enums.hpp
eseiler: then you should add `none = 0` :)

eseiler (changes requested):

include/method_enums.hpp
Irallia: Sorry, I expressed myself badly. I can't run the algorithm without a detecting method, but without a refinement method. So here 0 simply does not exist but in the refinement there is no_refinement = 0.

eseiler: Well, 0 is the default, so it does exist: https://godbolt.org/z/WGqbMx

IMO, either you enumerate them `0-3` or add a semantic meaning to `0` (`none`).
If you never need to check for `none`, then you can just do `0-3`.

Do you need the numerical values somewhere?
If not, you could omit them altogether.

I would always prefer to use the enum values instead of the numbers, both as user and developer.

eseiler (changes requested):
I replied in the comment thread

include/method_enums.hpp
Irallia: I understand, then I shift to 0. I need the numbers because I want to allow the user to specify -m 1 or just -m cigar_string.

src/detect_breakends/junction_detection.cpp
```diff
@@ -289,7 +289,7 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
             for (uint8_t method : methods) {
                 switch (method)
                 {
-                    case 1: // Detect junctions from CIGAR string
+                    case 0: // Detect junctions from CIGAR string
```
eseiler, line 292: You could also do the switch statement with enums:

```cpp
            for (detecting_methods method : methods) {
                switch (method)
                {
                    case detecting_methods::cigar_string: // Detect junctions from CIGAR string
```

the advantage is the verbosity and you already have the enums anyway.

eseiler (approved):
LGTM,

you can think about using enums instead of uint8_t throughout the code for the methods

src/detect_breakends.cpp
eseiler: This should work once https://github.com/seqan/seqan3/pull/2394 is merged

src/detect_breakends/junction_detection.cpp
```diff
@@ -245,7 +245,7 @@ void analyze_cigar(std::string chromosome,
  */
```
eldariont, line 245: This is a duplicate of the documentation in the associated header file. If I remember correctly, Jörg said that it's sufficient to have documentation in the header file.

include/detect_breakends/junction_detection.hpp
```diff
@@ -35,7 +35,7 @@
  */
 void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_file_path,
                                         const std::filesystem::path & insertion_file_path,
-                                        const std::vector<uint8_t> methods,
+                                        const std::vector<detecting_methods> methods,
```
eldariont, line 38: The name `detecting_methods` does not sound right. Maybe rename it to `detection_methods` if it's not too much of a hassle?

include/detect_breakends/junction_detection.hpp
```diff
@@ -15,10 +15,10 @@
  *
  * \param alignment_file_path input file - path to the sam/bam file
```
eldariont, line 16: The usage of the `-` in the documentation comments should be made consistent. Some params have the `-` after the parameter name, others in the middle of the description and others do not have any `-`.
My suggestion:
`\param parameter_name - parameter description`

include/method_enums.hpp
eldariont: Besides the test that Enrico suggested I cannot think of other applications of SIZE (can you?). It does not hurt to include the size though. However, I would add the suggested test already to this PR.

Two technical questions for my understanding: Why does this SIZE work? Is it because it get the value 4 by default and thus returns the size? And is there no other way of getting the size of an enum without setting SIZE explicitly? 🙂

eldariont (commented):
Looks good to me. I have only a few small comments.
However, we should probably wait for https://github.com/seqan/seqan3/pull/2394.

include/method_enums.hpp
eseiler: > Besides the test that Enrico suggested I cannot think of other applications of SIZE (can you?). It does not hurt to include the size though. However, I would add the suggested test already to this PR.

I also don't see another application, but if the default shall be "all methods", then it would probably be good to test this somewhere. Seems like something that is easy to forget. There may be another application regarding allowing certain combinations, e.g. creating a table of allowed combinations (which would require a known size).

> Two technical questions for my understanding: Why does this SIZE work? Is it because it gets the value 4 by default and thus returns the size? And is there no other way of getting the size of an enum without setting SIZE explicitly? 🙂

`SIZE` only works because we do the standard enumeration (0,1,2,3,4,5,...); otherwise there is no (? at least no simple) way to get the number of entries in an enum.

src/detect_breakends.cpp
Irallia: @eseiler No its not 😞 
If you want to try, I added a debug_stream to check and you could test with:
```bash
lydia@alnus:~/build/iGenVar$ ./bin/detect_breakends ~/Repos/iGenVar/test/data/simulated.minimap2.hg19.coordsorted_cutoff.sam ~/Repos/iGenVar/test/data/out.fasta -m 1 -m 2
Methods to be used: [cigar_string,split_read,read_pairs,read_depth,split_read,read_pairs]
....
```

src/detect_breakends.cpp
```diff
@@ -126,7 +126,13 @@ int main(int argc, char ** argv)
         return -1;
     }
 
-    seqan3::debug_stream << "Methods to be used: " << args.methods << std::endl;
+    // Check method selection uses not more than all methods.
+    if (args.methods.size() > detecting_methods::SIZE)
+    {
+        seqan3::debug_stream << "[Error] More methods were selected than exist. There may be duplications.\n";
```
eseiler, line 132:
```suggestion
    // Check that method selection contains no duplicates.
    std::vector<detecting_methods> unique_methods{args.methods};
    std::ranges::sort(unique_methods);
    unique_methods.erase(std::unique(unique_methods.begin(), unique_methods.end()), unique_methods.end());
    if (args.methods.size() > unique_methods.size())
    {
        seqan3::debug_stream << "[Error] The same detection method was selected multiple times.\n";
```

Otherwise, you can have duplications as long as you do not select more than 4.

Irallia: thanks you are of course right, that was lazy programming 😇

Irallia:
@eldariont After a rebase, I saw that I missed some of your review comments. Now it should be fine.
I found another bug, for which I opened an issue: https://github.com/seqan/iGenVar/issues/78


> `\param parameter_name - parameter description`

And for the parameter description, I will open another PR.

test/cli/detect_breakends_cli_test.cpp
```diff
@@ -59,9 +59,14 @@ const std::string help_page_part_2
...
 const std::string help_page_advanced
 {
-    "    -m, --method (List of unsigned 8 bit integer)\n"
-    "          Choose the detecting method(s) to be used. Default: [1,2,3,4]. Value\n"
-    "          must be in range [1,4].\n"
+    "    -m, --method (List of detection_methods)\n"
+    "          Choose the detection method(s) to be used. Default:\n"
+    "          [cigar_string,split_read,read_pairs,read_depth]. Value must be one\n"
+    "          of\n"
+    "          [read_depth,read_depth,read_pairs,read_pairs,split_read,split_read,cigar_string,cigar_string].\n"
```
eldariont, line 66: Why is each method included twice?

test/api/junction_detection_methods_test.cpp
```diff
@@ -86,9 +86,9 @@ TEST(junction_detection, single_method_only)
 
 TEST(junction_detection, method_pairs)
 {
-    for (uint8_t method_i : all_methods)
+    for (detection_methods method_i : all_methods)
     {
-        for (uint8_t method_j : all_methods)
+        for (detection_methods method_j : all_methods)
         {
```
eldariont, line 92: I'm not completely happy with the structure of these tests checking different method combinations. Do we really need the nested for-loops when the if-statements below check each combination explicitly anyways? I think the test could be made a little simpler without the for-loops but it's not a big deal and we can leave it like it is.

eldariont (approved):
Thanks for addressing my comments. In my opinion, this PR can be merged.

test/cli/detect_breakends_cli_test.cpp
Irallia: We don't use the new Validator for the detection methods enum, because I got a bug in a test. I opened an issue which should solve this problem: https://github.com/seqan/iGenVar/issues/78.

test/api/junction_detection_methods_test.cpp
Irallia: Yes I understand what you mean, but I thought with this solution the tests will be a little grouped and created more automatically, which may avoid mistakes.
I would leave it like this, but of course you can always make a PR with better tests. :wink:

eldariont: Yes, I see your point and as the test does what it promises to do we can leave it as it is.

PR #65, opened 2021-02-16
[DOC] Add workflow

Irallia:
This figure shows the workflow from iGenVar.
iGenVar_Workflow
The green filled boxes are done. The dotted boxes may not be necessary or will follow later.

An older Version I had was: 
iGenVar_Workflow
Tell me, if you like something of it more and would like me to add it to the newer one.


joshuak94 (commented):
Looks good! I would however change the bam index diagram. There is an index for each input, so either two separate bam index images, or just merge it into the sam/bam file diagram and make it say "Short Reads SAM/indexed BAM" and "Long Reads SAM/indexed BAM".

Actually, I don't think we should be supporting SAM files, as they can't be indexed. Most of these detection/refinement algorithms rely on the indexing to jump around, so we should either not support this, or make it explicit that we are converting SAM files to BAM files, then sorting and indexing them behind the scenes. But I'd say it's easier to just require indexed and sorted BAM files, which is standard practice anyways.

eldariont (commented):
Thanks for drawing the workflow! I like it :)

I agree with the two points Josh made: We don't need to have the index as a separate entity and for the time being we can require indexed BAMs as input (adding convenience features such as SAM->BAM conversion can be done later).

1. I think it would be easy to already add the clustering method of SVIRL at this stage: "Vertex cover in conflict graph (SVIRL)". SVIRL's refinement method could be renamed to "Consensus computation and realignment (SVIRL)" to make it more explicit.
2. The tool names should be `SVIM` instead of `SVim` and `SVIRL` instead of `SVirl`.
3. This workflow depicts the modules but not the interfaces / data types between them. Do we want to add them? At this point, we have not clearly defined the interfaces between the modules so I'm not sure. What we already have is the `Breakend` class as output of the CIGAR and Split-read methods and some kind of `ReadPairEvidence` class as output of the Read-pair method..

Irallia:
>     3. This workflow depicts the modules but not the interfaces / data types between them. Do we want to add them? At this point, we have not clearly defined the interfaces between the modules so I'm not sure. What we already have is the `Breakend` class as output of the CIGAR and Split-read methods and some kind of `ReadPairEvidence` class as output of the Read-pair method..

Yes I think for now I would leave it like that. We could make the workflow wider later and write the interfaces in the purple arrows.

Irallia:
iGenVar_Workflow

joshuak94 (approved):
Just some small things:
1. If we want to be nitpicky, SViper also receives the long read BAM file, since it extracts and corrects the long reads based on the short reads which cover these breakpoints.
2. I'm not sure what your comment means with "Maybe move behind clustering" for the read depth evidence?

eldariont (approved):
Looks good to me.

About the two things that @joshuak94 pointed out:
1. Maybe it would be best to remove the connection between the short reads and SViper as no other connection is drawn explicitly.
2. @Irallia and I discussed the optimal application of read depth a few weeks ago. We agreed that read depth is a very imprecise type of information and that it might be better to use it as kind of a refinement step after clustering the other types of information (e.g. adding additional support to split-read clusters). Does that make sense?

Irallia:
iGenVar_Workflow

PR #75, opened 2021-02-26
Set ref_id to -1 and ref_pos to 0 if they are unset. Skip those reads.

joshuak94:
In the case that a read is unmapped, the ref_id/ref_pos could be empty. Previously, ref_id was set to 0, but 0 refers to the first chromosome, which isn't right.

Instead, we should skip those reads which have invalid positions (1-based coordinates) or IDs (0-based). 

I think this should update that!

### DIFF
--- a/src/detect_breakends/junction_detection.cpp
+++ b/src/detect_breakends/junction_detection.cpp
@@ -71,7 +71,7 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
     {
         const std::string query_name            = seqan3::get<seqan3::field::id>(rec);                      // 1: QNAME
         const seqan3::sam_flag flag             = seqan3::get<seqan3::field::flag>(rec);                    // 2: FLAG
-        const int32_t ref_id                    = seqan3::get<seqan3::field::ref_id>(rec).value_or(0);      // 3: RNAME
+        const int32_t ref_id                    = seqan3::get<seqan3::field::ref_id>(rec).value_or(-1);     // 3: RNAME
         const int32_t pos                       = seqan3::get<seqan3::field::ref_offset>(rec).value_or(0);  // 4: POS
         const uint8_t mapq                      = seqan3::get<seqan3::field::mapq>(rec);                    // 5: MAPQ
         std::vector<seqan3::cigar> cigar        = seqan3::get<seqan3::field::cigar>(rec);                   // 6: CIGAR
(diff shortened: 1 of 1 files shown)


src/detect_breakends/junction_detection.cpp
```diff
@@ -283,67 +283,64 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
     {
         const std::string query_name            = seqan3::get<seqan3::field::id>(rec);                      // 1: QNAME
         const seqan3::sam_flag flag             = seqan3::get<seqan3::field::flag>(rec);                    // 2: FLAG
-        const int32_t ref_id                    = seqan3::get<seqan3::field::ref_id>(rec).value_or(0);      // 3: RNAME
+        const int32_t ref_id                    = seqan3::get<seqan3::field::ref_id>(rec).value_or(-1);     // 3: RNAME
         const int32_t pos                       = seqan3::get<seqan3::field::ref_offset>(rec).value_or(0);  // 4: POS
```
eldariont, line 287: According to the [seqan docs](https://docs.seqan.de/seqan/3-master-dev/group__io.html#ga2511b22375848e1bedbc5013c952a817) both ref_id and pos are 0-based. Therefore, you can also set a missing pos to -1:

```suggestion
        const int32_t pos                       = seqan3::get<seqan3::field::ref_offset>(rec).value_or(-1);  // 4: POS
```

src/detect_breakends/junction_detection.cpp
```diff
@@ -283,67 +283,64 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
...
-                                      seq,
-                                      junctions,
-                                      insertion_alleles,
-                                      min_var_length,
-                                      insertion_file);
-                        break;
-                    case 2: // Detect junctions from split read evidence (SA tag, primary alignments only)
-                        if (!hasFlagSupplementary(flag))
+        if (hasFlagUnmapped(flag) || hasFlagSecondary(flag) || hasFlagDuplicate(flag) || mapq < 20 ||
+            ref_id < 0 || pos < 1)
```
eldariont, line 296: Same here:
```suggestion
        if (hasFlagUnmapped(flag) || hasFlagSecondary(flag) || hasFlagDuplicate(flag) || mapq < 20 ||
            ref_id < 0 || pos < 0)
```

eldariont (changes requested):
Looks good. Just two little suggestions based on what we discovered yesterday (the fact that reference positions are 0-based in seqan). I will create a separate issue for that later but as you are already modifying this code, you could just fix it now :)

eldariont (approved):
Thanks!

joshuak94:
@Irallia Can you merge this one?

PR #79, opened 2021-03-06
[FEATURE] modularisation input

### LINKED ISSUE #43: [FEATURE] Modularize input data

### LINKED ISSUE #44: iGenVar - Modularisation of iGenVar
We want to modularise the different parts of IGenVar so that the user can decide which methods to use and so that we can compare different combinations of methods more easily. This also gives us the possibility to easily add a module to IGenVar, e.g. a new cluster method.

This affects
- [ ] Input Data
- [x] methods for detecting junctions
- [x] methods for clustering junctions
- [x] methods for refining breakends
- [x] move everything into different directories

Irallia:
Resolves #43 
This PR handles the last open ToDo of the modularisation epic and thus resolves #44.

Now we distinguish long and short read input data and can handle them separately.

### DIFF
--- a/test/api/junction_detection_methods_test.cpp
+++ b/test/api/junction_detection_methods_test.cpp
@@ -4,6 +4,8 @@
 
 #include "variant_detection/variant_detection.hpp"
 
+const std::string empty_alignment_short_reads_file_path = "";
+const std::string default_alignment_long_reads_file_path = DATADIR"simulated.minimap2.hg19.coordsorted_cutoff.sam";
 std::filesystem::path tmp_dir = std::filesystem::temp_directory_path();     // get the temp directory
 std::filesystem::path empty_path{};
 const std::vector<detection_methods> all_methods{cigar_string, split_read, read_pairs, read_depth};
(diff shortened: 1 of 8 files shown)


eaasna (changes requested):
Very nice work! 

Just a typo

eaasna (approved):

test/api/junction_detection_methods_test.cpp
```diff
@@ -67,13 +69,14 @@ TEST(junction_detection, single_method_only)
...
                              << "Test Method: " << method << '\n'
                              << "-----------------------------------------------------------------------\n";
-        detect_variants_in_alignment_file(DATADIR"simulated.minimap2.hg19.coordsorted_cutoff.sam",
-                                           tmp_dir/"detect_breakends_out_short.fasta",
-                                           {method},
-                                           simple_clustering,
-                                           no_refinement,
-                                           sv_default_length,
-                                           empty_path);
+        detect_variants_in_alignment_file(empty_alignment_short_reads_file_path,
```
eldariont, line 72: With the new `detect_junctions_in_long_reads_sam_file()` method that only performs junction detection (no clustering, refinement nor output) we are in a good position now to transform these tests into real unit tests. Instead of calling `detect_variants_in_alignment_file()` here, we could call `detect_junctions_in_long_reads_sam_file()` and check the resulting junction vector.

One option would be to convert each junction in the output vector to a string (using the stream operator) and compare with the strings in line 19 and 29. Alternatively, we could define a vector of expected junctions that we compare with the actual output vector from `detect_junctions_in_long_reads_sam_file()`.

You can decide whether you want to do this in this PR or whether I shall do it later in a separate PR.

eldariont: I saw just now that you are doing exactly what I just wrote in the new test at the end of `test/api/junction_detection_test.cpp`. Now the question is in which files to put the different tests. In my opinion, the new test could go into `test/api/junction_detection_methods_test.cpp`because it is testing the junction detection. We can add multiple similar tests to check the output of the different detection methods.

eldariont (changes requested):
Thanks for this great PR. I like how it pulls the junction detection out of the `detect_variants_in_alignment_file()` method which makes testing it much easier. I have only a few smaller comments and a suggestion about the tests.

test/api/junction_detection_methods_test.cpp
Irallia: Let me think for a moment. 🤔 

We want to split our `detect_variants_in_alignment_file()` so that we have methods, clustering and refinement separately.
That means we then have something like `detection_methods_test`, `clustering_test` and `refinement_test`.
Furthermore we distinguish between short and long reads, which leads us to `input_test` or similar? Or the above tests include both as possible inputs. I think I prefer the latter.

The `junctiond_detection_test.cpp` currently includes a mix of clustering and refinement and I had moved the methods to the separate file `junction_detection_methods_test.cpp`.

So, yes I will move the test to the other file and then we should sort the rest when we change `detect_variants_in_alignment_file()`.

Thanks for pointing this out. 😃

eldariont (approved):
Thanks for implementing the suggested chances. Looks good to merge now 😃

PR #80, opened 2021-03-09
Add random sampling for read depth estimation.

joshuak94:
This pull request adds the misc subfolder in the include directory. For now it only has the header file for calculating the random positions for sampling. I'm open to suggestions on the organization of this!

Other than that, I created a preliminary random position generator which generates random chromosomes and positions. It isn't perfect, because it's based off of the standard library. But for an initial implementation it should suffice.

The random positions are basically pre-calculated and sorted, and then the depth for these positions are calculated on the fly as the reads overlap them. This assumes the BAM file is of course sorted.


joshuak94:
Eventually I'd like to implement estimations for insert length for paired end reads and read lengths in general.

src/detect_breakends/junction_detection.cpp
```diff
@@ -5,6 +5,7 @@
 #include <seqan3/io/sequence_file/output.hpp>   // FASTA support
 
 #include "detect_breakends/bam_functions.hpp"                       // for hasFlag* functions
+#include "misc/statistics.hpp"                                      // for calculating average read depth
```
Irallia, line 8: Okay, this means, we would have
- /modules/METHOD_NAME/ for all methods
- /structures/ for our sturcures like junctions, breakends,....
- /variant_detections/
- /variant_parser/
- /misc/ for general functions?
We should discuss this in the todays meeting.

Irallia (commented):
I have already looked over it once. But I think this is a PR that can have as 1st review someone from the seqan team.

src/detect_breakends/junction_detection.cpp
```diff
@@ -57,6 +66,26 @@ void detect_junctions_in_alignment_file(const std::filesystem::path & alignment_
...
+            else if (ref_id > std::get<0>(rand_sample.front()) ||       // Record is at next chromosome
+                     (ref_id == std::get<0>(rand_sample.front()) &&     // Record is at same chromosome BUT
+                      pos > std::get<1>(rand_sample.front())))          // in a later position.
+            {
+                ++num_pos;
+                avg_depth += cur_depth;
+                cur_depth = 0;
+                rand_sample.pop_front();
+            }
+        }
```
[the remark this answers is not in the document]
joshuak94: I'm not sure if it's such a good idea to put this into a separate function. I mean for one, it's basically just three if statements. The other thing is that I'm computing the depth at the current position, and then when we move to the next position this gets added to the average and then the current depth is zeroed out. So I'd have to pass all of these variables in and I think it would just get more confusing jumping back and forth between functions.

include/misc/statistics.hpp
```diff
@@ -0,0 +1,53 @@
...
+/*! \brief Returns a list of pairs, where the first integer is a random chromosome and the second is a random position.
+ *        The list is sorted by default.
+ *  \param sample_size - The number of positions to generate.
+ *  \param header - The header from the SAM file input.
+ * 
+ *  \return Returns a sorted forward_list of pairs of reference chromosome IDs and positions.
+ */
+std::forward_list<std::pair<int32_t, int32_t>> get_random_positions(uint64_t sample_size, seqan3::sam_file_header<std::deque<std::string>> & header)
+{
+    std::forward_list<std::pair<int32_t, int32_t>> output{};
```
marehr, line 16: Is this performance critical? I wouldn't use forward lists otherwise.

include/misc/statistics.hpp
```diff
@@ -0,0 +1,53 @@
...
+    if (sample_size == 0)
+    {
+        output.push_front(std::pair<int32_t, int32_t>{-1, -1});
```
marehr, line 19:
```suggestion
        output.push_front(std::pair<int32_t, int32_t>{-1, -1});
        return output;
```

this makes it clear that this is a special case.

include/misc/statistics.hpp
marehr:
```suggestion
    std::vector<std::pair<int32_t, int32_t>> random_positions(sample_size); // pre-allocate sample-size many
```

include/misc/statistics.hpp
```diff
@@ -0,0 +1,53 @@
...
+    int32_t rand_chr{};
+    int32_t rand_pos{};
+
+    for (int i = 0; i < sample_size; ++i)
+    {
+        rand_chr = std::rand() % header.ref_ids().size();
+        rand_pos = std::rand() % std::get<0>(header.ref_id_info[rand_chr]);
+
+        output.push_front(std::pair<int32_t, int32_t>{rand_chr, rand_pos});
+    }
```
marehr, line 30:
```suggestion
```suggestion
    std::ranges::generate(random_positions, [&header]()
    {
        int32_t rand_chr = std::rand() % header.ref_ids().size();
        int32_t rand_pos = std::rand() % std::get<0>(header.ref_id_info[rand_chr]);
        return std::pair<int32_t, int32_t>{rand_chr, rand_pos});
    });
```
```

File: include/misc/statistics.hpp [hunk 6]
```diff
@@ -0,0 +1,53 @@
...
+    if (sample_size == 0)
+    {
+        output.push_front(std::pair<int32_t, int32_t>{-1, -1});
+    }
+    int32_t rand_chr{};
+    int32_t rand_pos{};
+
+    for (int i = 0; i < sample_size; ++i)
+    {
+        rand_chr = std::rand() % header.ref_ids().size();
```
marehr, line 26: You should consider using the "real" random module. 

https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution#Example

That would also make clear what your distribution is.

Furthermore, I would require passing in a random generator into this function so that the "randomness" can be controlled from the outside.

include/misc/statistics.hpp
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+uint32_t get_read_length(std::vector<seqan3::cigar> & cigar)
+{
+    using seqan3::get;
+
+    uint32_t result{0};
+    for (auto e : cigar)
+    {
+        result += get<0>(e);
```
marehr, line 49: Is this correct? 

The sam specification says:

| Op | BAM | Description | Consumes query | Consumes reference |
| -- | --- | ----------- | -------------- | ------------------ |
| M  | 0   | alignment match (can be a sequence match or mismatch) | yes | yes |
| I  | 1   | insertion to the reference |yes | no |
| D  | 2   | deletion from the reference | no | yes |
| N  | 3   | skipped region from the reference | no | yes |
| S  | 4   | soft clipping (clipped sequences present in SEQ) | yes | no |
| H  | 5   | hard clipping (clipped sequences NOT present in SEQ) | no | no |
| P  | 6   | padding (silent deletion from padded reference) | no | no |
| =  | 7   | sequence match | yes | yes |
| X  | 8   | sequence mismatch | yes | yes |

“Consumes query” and “consumes reference” indicate whether the CIGAR operation causes the alignment to step along the query sequence and the reference sequence respectively.

So I would say that you over-count the read length.

https://samtools.github.io/hts-specs/SAMv1.pdf#page=8

src/variant_detection/variant_detection.cpp
```diff
@@ -60,6 +68,26 @@ void detect_variants_in_alignment_file(const std::filesystem::path & alignment_f
             continue;
 
         const std::string ref_name = ref_ids[ref_id];
+        uint32_t read_length{get_read_length(cigar)};
+        // If there are positions to be sampled, then sample them.
+        if (!rand_sample.empty() && sample_size > 0)
+        {
+            if (ref_id == std::get<0>(rand_sample.front()) &&           // Correct chromosome.
```
marehr, line 75:
```suggestion
            auto && [chr, pod] = rand_sample.front();
            if (ref_id == chr) &&           // Correct chromosome.
```

give the pair content a name that makes it easier to understand.

src/variant_detection/variant_detection.cpp
```diff
@@ -60,6 +68,26 @@ void detect_variants_in_alignment_file(const std::filesystem::path & alignment_f
...
+                pos <= std::get<1>(rand_sample.front()) &&              // Read starts before or at the position.
```
marehr, line 76: again, not all cigar length's are for reads, they can also only affect the sequence.

src/variant_detection/variant_detection.cpp
```diff
@@ -60,6 +68,26 @@ void detect_variants_in_alignment_file(const std::filesystem::path & alignment_f
...
+                ++cur_depth;
+            }
+            else if (ref_id > std::get<0>(rand_sample.front()) ||       // Record is at next chromosome
+                     (ref_id == std::get<0>(rand_sample.front()) &&     // Record is at same chromosome BUT
+                      pos > std::get<1>(rand_sample.front())))          // in a later position.
+            {
+                ++num_pos;
+                avg_depth += cur_depth;
+                cur_depth = 0;
+                rand_sample.pop_front();
```
marehr, line 88: I'm not completely sure what this does and why you consume your "random" sample throughout the loop without "refreshing" it.

What happens if your random sample doesn't get fully consumed, but your next reference starts? Is that possible?

marehr (commented):
Some remarks, no hard feeling, you can also just ignore it :)

src/detect_breakends/junction_detection.cpp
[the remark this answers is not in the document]
marehr: My general feeling is that decoupling is always better and "just" 3 or 4 or 5 or 20 if branches can change pretty fast.

Issue #77, opened 2021-03-04
[INFRA] Combine two existing executables into a single monolithic executable


eldariont:
**Original idea (2019):**
- Implementation of caller modules as separate executables
- Combination of modules with unix pipes, e.g. `./detect_breakends.cpp | ./find_deletions.cpp > output.vcf`

**Current plan (2021):**
- Implementation of caller modules in separate source files that are combined into a single large executable

**Therefore, we need to**
- Remove `find_deletions.cpp` executable and integrate code into the main executable (currently called `detect_breakends.cpp`)
- Update all associated tests
- Rename `detect_breakends.cpp` into something more generic like `iGenVar.cpp`

Irallia:
Resolved by #81

PR #81, opened 2021-03-09
[INFRA] Combine executables

eldariont:
This is my draft of a fix for #77.

This PR:
- removes the find_deletions.cpp executable and integrates its code (mainly the VCF output) into the main executable (previously called detect_breakends.cpp)
- renames detect_breakends.cpp to iGenVar.cpp because it is a better name for the final tool
- updates all associated tests

With this PR, iGenvar now 1) detects junctions, 2) clusters them, 3) detects deletions based on the junctions, and 4) outputs the deletions to a VCF file or stdout. In future PRs, I can add the detection of insertions, translocations and other variants based on the junctions which should be very straightforward.

As all existing tests check the output on stdout or stderr, this PR broke many of them. The used test input file `simulated.minimap2.hg19.coordsorted_cutoff.sam` contains translocations and an insertion (no deletions) which is why the VCF output is currently empty. I updated the tests to expect an empty VCF output so that they now pass. However, this makes the tests relatively pointless and we need to replace them with proper API tests that test each method separately.


src/iGenVar.cpp
```diff
@@ -59,20 +59,21 @@ struct cmd_arguments
...
 void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments & args)
 {
-    parser.info.author = "David Heller & Lydia Buntrock";
+    parser.info.author = "Lydia Buntrock, David Heller, Joshua Kim";
     parser.info.app_name = "iGenVar";
-    parser.info.man_page_title = "Short and long Read SV Caller";
-    parser.info.short_description = "Detect junctions in a read alignment file";
+    parser.info.man_page_title = "Short and Long Read SV Caller";
+    parser.info.short_description = "Detect genomic variants in a read alignment file";
     parser.info.version = "0.0.1";
```
Irallia, line 75: Should we increase the version at certain steps? Maybe when the modularisation and the big infra stuff is done?

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -1,6 +1,6 @@
 #include <seqan3/core/debug_stream.hpp>
 
-#include "detect_breakends/bam_functions.hpp"   // for hasFlag* functions
+#include "variant_detection/bam_functions.hpp"   // for hasFlag* functions
```
Irallia, line 3:
```suggestion
#include "variant_detection/bam_functions.hpp"  // for hasFlag* functions
```

test/api/junction_detection_methods_test.cpp
```diff
@@ -56,28 +43,29 @@ TEST(junction_detection, single_method_only)
...
                                            {method},
                                            simple_clustering,
                                            no_refinement,
-                                           sv_default_length);
+                                           sv_default_length,
+                                           empty_path);
 
         if (method == 0)
         {
-            check_output_and_cleanup(expected_res_cigar);
```
Irallia, line 68: Instead of replacing the line, could you (comment out the previous one and) add a todo (David) so we can see that something still needs to happen here and who added the todo?

Irallia (commented):
It all looks very good. 
It feels a bit strange to empty all the tests, but we had discussed that. Maybe we should mark with todos where we actually expect other results, which then still needs to be done.
Also, Josh should definitely look over the finished PR, since he did the VCF stuff and he can see better if it all still fits.

src/iGenVar.cpp
eldariont: Yes, good idea. Let's maybe wait for #81 and #82 until we move to 0.0.2?

Irallia: Jepp and maybe https://github.com/seqan/iGenVar/pull/79 aswell.

Irallia (approved):
Thank you for the TODO comments.
I think this is ready to review, but you can think about waiting after the other big PR is merged.
Whichever is more to your liking.

src/variant_detection/variant_output.cpp
```diff
@@ -0,0 +1,83 @@
...
+                        variant_record tmp{};
+                        tmp.set_chrom(mate1.seq_name);
+                        tmp.set_qual(60);
+                        tmp.set_alt("<DEL>");
+                        tmp.add_info("SVTYPE", "DEL");
+                        tmp.set_pos(mate1_pos);
+                        tmp.add_info("SVLEN", std::to_string(-distance));
+                        tmp.add_info("END", std::to_string(mate2_pos));
+                        tmp.print(out_stream);
+                        // print_deletion(mate1.seq_name, mate1_pos, mate2_pos, 60, out_stream);
```
joshuak94, line 40:
```suggestion

```

Can we delete the commented out function here and below?

joshuak94 (commented):
Looks good! Just some small suggestions...

eldariont:
> Looks good! Just some small suggestions...

Thanks for the good suggestions. I just added them.

include/variant_detection/variant_detection.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * \details Detects novel junctions from read alignment records using different detection methods.
+ *          The junctions are clustered using one of several clustering methods.
+ *          Then, the junction clusters are refined using one of several refinement methods.
+ *          Finally, the refined junction clusters are categorized into different variant classes
+ *          and output in VCF format.
+ */
+void detect_variants_in_alignment_file(const std::filesystem::path & alignment_file_path,
+                                       const std::filesystem::path & insertion_file_path,
+                                       const std::vector<detection_methods> methods,
```
joshuak94, line 35: Small point: Is there a reason this is passed by value and not by reference?

include/variant_detection/variant_output.hpp
```diff
@@ -0,0 +1,29 @@
...
+
+
+/*! \brief Detects genomic variants from junction clusters and prints them to output stream in VCF format.
+ *
+ * \param clusters input junction clusters
+ * \param out_stream output stream
+ *
+ * \details Extracts genomic variants from given junction clusters.
+ */
+void find_and_output_variant(std::vector<Cluster> const clusters, std::ostream & out_stream);
```
joshuak94, line 17: Here and below as well, why pass by value instead of pass by reference?

src/variant_detection/variant_output.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+void find_and_output_variant(std::vector<Cluster> const clusters, std::ostream & out_stream)
+{
+    variant_header header{};
+    header.set_fileformat("VCFv4.3");
+    header.add_meta_info("SVTYPE", 1, "String", "Type of SV called.", "iGenVarCaller", "1.0");
+    header.add_meta_info("SVLEN", 1, "Integer", "Length of SV called.", "iGenVarCaller", "1.0");
+    header.add_meta_info("END", 1, "Integer", "End position of SV called.", "iGenVarCaller", "1.0");
+    header.print(out_stream);
+    for (size_t i = 0; i<clusters.size(); ++i)
```
joshuak94, line 16:
```suggestion
    for (size_t i = 0; i < clusters.size(); ++i)
```
Just a small spacing consistency.

joshuak94 (changes requested):
Sorry I found a few more things.

include/variant_detection/variant_detection.hpp
eldariont: Good catch, this originates from earlier PRs but I don't see a reason why this needs to be by value.

joshuak94 (approved):
@Irallia looks good to me!

PR #82, opened 2021-03-10
[MISC] refactor representation of insertions

### LINKED ISSUE #76: [MISC] Refactor representation of insertions
**Current implementation**

A breakend is a directed position on the reference genome or a read, e.g. chr1, position 1000, positive strand or read 7:position 210, negative strand.
A novel adjacency is a pair of breakends and represents the connection of two distant positions on the genome or a read.

Deletions can be represented as adjacencies between distant positions on the same chromosome, e.g. chr1:1000 -> chr1: 2000.
Insertions can be represented as two adjacencies:
1. Adjacency 1 between the insertion location and the start of the insertion sequence on the read, e.g. chr1:1000 -> read7:210
2. Adjacency 2 between the end of the insertion sequence on the read and the insertion location, e.g. read7:310 -> chr1:1001

This representation of insertions is problematic because:
- A single event (the insertion) is represented by two novel adjacencies.
- We need both adjacencies to understand what is going on.
- Insertions are hard to cluster because we would need to merge adjacencies to different reads (e.g. chr1:1000 -> read7:210 and chr1:1000 -> read8:430 could come from the same insertion).

**Desired implementation**

A breakend is a directed position on the reference genome, e.g. chr1, position 1000, positive strand.
A novel adjacency is a pair of breakends and represents the connection of two distant genomic positions. Each novel adjacency has a field `insertion_sequence` that (optionally) stores additional bases inserted between the two joined genomic positions.
As before, deletions can be represented as adjacencies between distant positions on the same chromosome, e.g. chr1:1000 -> chr1: 2000.
Now, insertions can be represented as adjacencies between two neighboring genomic positions, e.g. chr1:1000 -> chr1:1001, with the insertion sequence stored in the respective field.

eldariont:
This PR fixes #76 (please see there for a detailed explanation).

This PR:
- removes field `sequence_type` from breakend class
- adds field `inserted_sequence` to junction class
- adds `get_average_inserted_sequence_size()` method to cluster class
- gets rid of output file insertion_file (seqs now stored in junction class)
- updates tests


src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -52,27 +52,27 @@ void retrieve_aligned_segments(std::string sa_string, std::vector<AlignedSegment
...
 {
     for(size_t i = 1; i<aligned_segments.size(); i++)
     {
         AlignedSegment current = aligned_segments[i-1];
         AlignedSegment next = aligned_segments[i];
         int32_t distance_on_read = next.get_query_start() - current.get_query_end();
-        // Neither gap nor overlap on read
-        if (distance_on_read >= -10 && distance_on_read <= 10)
+        // Check that there is neither an overlap nor a larger gap on the read
+        if (distance_on_read >= 0 && distance_on_read <= 10)
```
Irallia, line 64: Nothing about this PR but:
Where does this 10 come from. Is this a parameter which is normally 10 or do we want to give the opportunity to change it?

include/structures/junction.hpp
```diff
@@ -1,12 +1,18 @@
...
+#include <seqan3/range/views/to.hpp>
+
 #include "structures/breakend.hpp"
 
 class Junction
 {
 private:
     Breakend mate1{};
     Breakend mate2{};
+    seqan3::dna5_vector inserted_sequence{};
```
Irallia, line 15: I think we should update the equality function so that a junction is only equal if its inserted_sequence is also equal (or has a certain minimum edit distance?).
EDIT: You decided the are equal and we save all different insertions in a vector in the cluster right? We should maybe add this as a comment to the equality function.

include/structures/cluster.hpp
```diff
@@ -30,13 +30,16 @@ class Cluster
     Breakend get_average_mate1() const;
 
     Breakend get_average_mate2() const;
+
+    int32_t get_average_inserted_sequence_size() const;
```
Irallia, line 34: Some documentation?

Irallia (changes requested):
Thanks for the nice work.
I just have some minor comments.

include/modules/sv_detection_methods/analyze_sa_tag_method.hpp
```diff
@@ -57,5 +59,6 @@ void analyze_sa_tag(const std::string & query_name,
                     const int32_t pos,
                     const uint8_t mapq,
                     std::vector<seqan3::cigar> & cigar,
+                    const seqan3::dna5_vector seq,
                     const std::string sa_tag,
```
Irallia, line 63: Update documentation..

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
eldariont: This 10 is a totally arbitrary cutoff and in the end the user should get the opportunity to change it. When working on this PR I decided to leave this for later.
Shall I add a TODO or would you like to add a command-line option with this PR?

include/structures/junction.hpp
eldariont: Good point. I'll add the comparison of the inserted sequence to the comparison operators.
I'm not entirely sure I understand your second sentence. What do you suggest we add as a comment?

Irallia: Sorry for the confusion.
I just thought. We may want insertions with unequal sequences to be equal if the breakends are equal. Or maybe with similar sequences. But for now, maybe that's enough. But if you think we should add this later, you could add a TODO to the operator==() function.

include/structures/junction.hpp
```diff
@@ -19,26 +25,39 @@ class Junction
...
+    //!\}
 
+    // \brief Returns the first mate of this junction.
     Breakend get_mate1() const;
 
+    // \brief Returns the second mate of this junction.
     Breakend get_mate2() const;
 
+    /*! \brief Returns the inserted sequence inserted between the two two mates.
+    *          If the two mates are connected directly, the inserted sequence is empty.
```
Irallia, line 56:
```suggestion
    /*! \brief Returns the sequence inserted between the two two mates.
    *          If the two mates are connected directly, the inserted sequence is empty.
```

Irallia (changes requested):
Thanks for the great effort to catch up with all the documentation! 👍 
Found just one tiny documentation thing and there is the open question about similar sequences.

If we say, that junctions are equal only if the inserted sequences are equal, it would affect the clustering method. (We would cluster less.) The question is, do we want to cluster junctions with similar inserted sequences, and do we want to change this in the operator==() function, or do we add another function? If yes, we should open a new issue and add a TODO note here.

include/structures/junction.hpp
eldariont: Thanks for the explanation, @Irallia. I now see what you mean and it is a very good question.
In my mind, each junction stands for an operation on the genome and the complete set of junctions should, in theory, transform the reference genome into the patient genome. Two junctions that have identical breakends but different insertion sequences are not identical because they yield different genome sequences when applied to the reference genome. Therefore, I would prefer the operator==() function to implement a strict identity check.

You are right that this affects our (simple) clustering method and leads to very small clusters because only identical junctions get clustered together. However, I think this is not the problem of the operator==() function (which I would expect to be strict). Instead, it is because the simple clustering method uses the operator==() function and we should soon implement a clustering method that makes more sense. Like you write, we would like to cluster junctions with similar breakends and similar insertion sequences (not only identical junctions). But in my opinion this should be done in a better clustering function, not in the operator==() function.

eldariont:
Thanks for the review, @Irallia! I fixed the documentation typo.

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
Irallia: No a TODO and or an extra issue is good. 👍

eldariont: Great, I added a TODO with 4ebeccee6e1ce4351e9c8699500b1c2d1b4fa124.

Irallia (approved):
Very good, since this is so big and relevant in content I would pass josh the 2 review.
But you should rebase first.

eldariont:
Hi @joshuak94, I just rebased this PR to the current master so everything is in order now. It's a pretty big one so it would be nice if we could merge it sometime soon :)

include/modules/sv_detection_methods/analyze_cigar_method.hpp
```diff
@@ -35,6 +33,4 @@ void analyze_cigar(const std::string & read_name,
                    std::vector<seqan3::cigar> & cigar_string,
                    const seqan3::dna5_vector & query_sequence,
                    std::vector<Junction> & junctions,
-                   std::vector<seqan3::dna5_vector> & insertions,
-                   int32_t min_length,
-                   seqan3::sequence_file_output<> & insertion_file);
+                   int32_t min_length);
```
joshuak94, line 36: Can this be const?

include/modules/sv_detection_methods/analyze_sa_tag_method.hpp
```diff
@@ -57,5 +60,6 @@ void analyze_sa_tag(const std::string & query_name,
                     const int32_t pos,
                     const uint8_t mapq,
                     std::vector<seqan3::cigar> & cigar,
+                    const seqan3::dna5_vector seq,
```
joshuak94, line 63: Can this be passed by reference?

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -52,27 +52,28 @@ void retrieve_aligned_segments(std::string sa_string, std::vector<AlignedSegment
...
-                            next.orientation == strand::forward ? next.get_reference_start()
-                                                                : next.get_reference_end(),
-                            next.orientation,
-                            sequence_type::reference};
-            Junction new_junction{std::move(mate1), std::move(mate2), read_name};
+                           next.orientation == strand::forward ? next.get_reference_start()
+                                                               : next.get_reference_end(),
+                           next.orientation};
+            auto inserted_bases = query_sequence | seqan3::views::slice(current.get_query_end(), next.get_query_start());
+            Junction new_junction{std::move(mate1), std::move(mate2), std::move(inserted_bases), read_name};
```
joshuak94, line 76: Do you need to do std::move here? Isn't there a std::move in the constructor as well?

src/structures/breakend.cpp
```diff
@@ -2,19 +2,16 @@
...
-            : lhs.seq_name != rhs.seq_name
-                ? lhs.seq_name < rhs.seq_name
-                : lhs.orientation != rhs.orientation
-                    ? lhs.orientation < rhs.orientation
-                    : lhs.position < rhs.position;
+    return lhs.seq_name != rhs.seq_name
+            ? lhs.seq_name < rhs.seq_name
+            : lhs.orientation != rhs.orientation
+                ? lhs.orientation < rhs.orientation
+                : lhs.position < rhs.position;
```
joshuak94, line 9: This is such a weird looking nested ternary... But I assume it's correct!

src/structures/cluster.cpp
```diff
@@ -54,24 +50,34 @@ Breakend Cluster::get_average_mate2() const
...
-    Breakend average_breakend{seq_name, average_position, orientation, seq_type};
+    Breakend average_breakend{seq_name, average_position, orientation};
     return average_breakend;
 }
+
+int32_t Cluster::get_average_inserted_sequence_size() const
+{
+    uint64_t sum_sizes = 0;
+    // Iterate through members of the cluster
+    for (uint32_t i = 0; i < members.size(); ++i)
```
joshuak94, line 77:
```suggestion
    for (size_t i = 0; i < members.size(); ++i)
```

joshuak94 (changes requested):
Sorry it took me so long! @eldariont Looks good, just some small things!

include/modules/sv_detection_methods/analyze_cigar_method.hpp
eldariont: Of course, thanks for spotting this.

include/modules/sv_detection_methods/analyze_sa_tag_method.hpp
eldariont: Yes indeed.

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
eldariont: Thanks for this comment. It turned out I just was not aware of what `std::move` actually does. You are completely right and I remove the `std::move` around `mate1`, `mate2` and `inserted_bases`.
I also made the `inserted_sequence` parameter in the `Junction` constructor const reference because it is only used as an input to a pipeline of views.

src/structures/breakend.cpp
joshuak94: I think a short comment explaining would be fine!

joshuak94 (approved):

PR #89, opened 2021-03-18
[DOC] unify param description

Irallia:
For a more uniform documentation, we have established a few rules that can be read in https://github.com/seqan/iGenVar/wiki/Coding-Guide. I have now adapted the existing documentation for this purpose.

### DIFF
--- a/include/modules/sv_detection_methods/analyze_cigar_method.hpp
+++ b/include/modules/sv_detection_methods/analyze_cigar_method.hpp
@@ -2,13 +2,13 @@
 
 /*! \brief This function steps through the CIGAR string and stores junctions with their position in reference and read.
  *
- * \param read_name         - QNAME field of the SAM/BAM file
- * \param chromosome        - RNAME field of the SAM/BAM file
- * \param query_start_pos   - POS field of the SAM/BAM file
- * \param cigar_string      - CIGAR field of the SAM/BAM file
- * \param query_sequence    - SEQ field of the SAM/BAM file
- * \param junctions         - vector for storing junctions
- * \param min_length        - minimum length of variants to detect (default 30 bp)
+ * \param[in]       read_name       - QNAME field of the SAM/BAM file
+ * \param[in]       chromosome      - RNAME field of the SAM/BAM file
+ * \param[in]       query_start_pos - POS field of the SAM/BAM file
+ * \param[in]       cigar_string    - CIGAR field of the SAM/BAM file
+ * \param[in]       query_sequence  - SEQ field of the SAM/BAM file
+ * \param[in, out]  junctions       - vector for storing junctions
+ * \param[in]       min_length      - minimum length of variants to detect (default 30 bp)
  *
  * \details This function steps through the CIGAR string and stores junctions with their position in reference and read.
  *          We distinguish 4 cases of CIGAR operation characters:
--- a/include/modules/sv_detection_methods/analyze_cigar_method.hpp
+++ b/include/modules/sv_detection_methods/analyze_cigar_method.hpp
@@ -2,13 +2,13 @@
 
 /*! \brief This function steps through the CIGAR string and stores junctions with their position in reference and read.
  *
- * \param read_name         QNAME field of the SAM/BAM file
- * \param chromosome        RNAME field of the SAM/BAM file
- * \param query_start_pos   POS field of the SAM/BAM file
- * \param cigar_string      CIGAR field of the SAM/BAM file
- * \param query_sequence    SEQ field of the SAM/BAM file
- * \param junctions         vector for storing junctions
- * \param min_length        minimum length of variants to detect (default 30 bp)
+ * \param read_name         - QNAME field of the SAM/BAM file
+ * \param chromosome        - RNAME field of the SAM/BAM file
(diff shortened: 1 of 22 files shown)


SGSSGene (approved):
:cry: "west const"? It is a revolution https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
It is like using useless Emacs over clearly superior vim!

The "[in, out]" tags are great, and I like the " - " before the description of the variables. Makes them much better to read in code, and they look fine in html:

screen

Irallia:
> 😢 "west const"? It is a revolution https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
> It is like using useless Emacs over clearly superior vim!

Don't know. @eldariont @joshuak94 what do you say? We can also use west const. I just want to have a general solution ^^

joshuak94:
> > cry "west const"? It is a revolution https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
> > It is like using useless Emacs over clearly superior vim!
> 
> Don't know. @eldariont @joshuak94 what do you say? We can also use west const. I just want to have a general solution ^^

Hmmmm it seems a bit unnatural... But I like the blog post and what they're trying to accomplish so I'm fine with moving to east const...

joshuak94 (approved):
Looks good! Nice work with the style choices!

eldariont:
> > 😢 "west const"? It is a revolution https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
> > It is like using useless Emacs over clearly superior vim!
> 
> Don't know. @eldariont @joshuak94 what do you say? We can also use west const. I just want to have a general solution ^^

I don't have strong feeling about this either. The blog post was very interesting but I guess it would be totally fine either way. If we were an independent project I would pick east const. But if seqan uses west const I would stick to that for consistency.

Irallia:
> > > 😢 "west const"? It is a revolution https://mariusbancila.ro/blog/2018/11/23/join-the-east-const-revolution/
> > > It is like using useless Emacs over clearly superior vim!
> > 
> > Don't know. @eldariont @joshuak94 what do you say? We can also use west const. I just want to have a general solution ^^
> 
> I don't have strong feeling about this either. The blog post was very interesting but I guess it would be totally fine either way. If we were an independent project I would pick east const. But if seqan uses west const I would stick to that for consistency.

No, seqan uses east const. So, then let's break the egg at the little end and join the east const revolution. 😎
I checked the seqan3 wiki, let's follow these rules:
- when possible, make variable constexpr or const (in that order)
- always use "east-const", i.e. put const on the right of the type that is modified
- if a variable is constexpr, put the constexpr on the left of the type (west-constexpr)
see https://github.com/seqan/seqan3/wiki/General-formatting#variable

Irallia:
@SGSSGene @joshuak94 @eldariont  As I changed everything const related again, could someone recheck my changes? (last two commits) 😁

include/modules/sv_detection_methods/analyze_cigar_method.hpp
```diff
@@ -29,11 +29,11 @@
  *          For more information see the
  *          ([Map Format Specification](https://github.com/samtools/hts-specs/blob/master/SAMv1.pdf)) page 8.
  */
-void analyze_cigar(const std::string & read_name,
-                   const std::string chromosome,
-                   const int32_t query_start_pos,
+void analyze_cigar(std::string const & read_name,
+                   std::string const chromosome,
+                   int32_t query_start_pos,
```
eldariont, line 34: - can we pass `chromosome` by reference?
- was it intentional to drop the `const` of `query_start_pos`?

test/api/junction_detection_methods_test.cpp
```diff
@@ -253,10 +253,10 @@ TEST(junction_detection, detect_junctions_in_long_reads_sam_file)
...
     Breakend new_breakend_9 {chromosome_1, pos_read_4, strand::forward, sequence_type::reference};
 
-    const std::string & read_name_1 = "m2257/8161/CCS";
-    const std::string & read_name_2 = "m41327/11677/CCS";
-    const std::string & read_name_3 = "m21263/13017/CCS";
-    const std::string & read_name_4 = "m38637/7161/CCS";
+    std::string const & read_name_1 = "m2257/8161/CCS";
+    std::string const & read_name_2 = "m41327/11677/CCS";
+    std::string const & read_name_3 = "m21263/13017/CCS";
+    std::string const & read_name_4 = "m38637/7161/CCS";
```
SGSSGene, line 260: These shouldn't be references. (You could make them std::string_view)
```suggestion
    std::string const read_name_1 = "m2257/8161/CCS";
    std::string const read_name_2 = "m41327/11677/CCS";
    std::string const read_name_3 = "m21263/13017/CCS";
    std::string const read_name_4 = "m38637/7161/CCS";

```

SGSSGene (changes requested):
"East const" :metal:

There are some std::string references that shouldn't be.

include/modules/sv_detection_methods/analyze_cigar_method.hpp
Irallia: No, that was a mistake 😅

SGSSGene (approved):
nice :+1:

PR #99, opened 2021-04-08
[FEATURE] Add hierarchical clustering

### LINKED ISSUE #32: iGenVar - Call Deletions from long reads
**Call SVs:** For larger structural variations, we want to combine the various known methods to call deletions.
- ~~Distinguish between long and short reads in the input #17~~
    - [x] Call Deletions via the CIGAR string
        - [x] Call Deletions in long reads #25 
        - ~~Call Deletions in short reads #27~~
    - [x] Call Deletion using split alignment #18 
- [ ]  Cluster junctions by hierarchical clustering #54

----
EDIT (30.03.2021): We have removed some of the requirements as we are outsourcing them:
- short read input https://github.com/seqan/igenvar/issues/17

### LINKED ISSUE #54: [FEATURE] Cluster junctions by hierarchical clustering

eldariont:
This PR adds a second clustering method - agglomerative hierarchical clustering - and closes #54.

The idea
---------
Agglomerative hierarchical clustering uses a "bottom-up" approach: Each junction starts in its own cluster. In every iteration, the two clusters with the lowest distance are merged. This process yields a dendogram (see the example here[](https://en.wikipedia.org/wiki/Hierarchical_clustering#Agglomerative_clustering_example)) that can be cut at a desired distance.

To break the problem of clustering thousands of junctions into multiple smaller problems, we first generate smaller partitions of junctions in close proximity (see `partition_junctions()` in `src/modules/clustering/hierarchical_clustering_method.cpp`).

Then, we cluster each partition separately (see `hierarchical_clustering_method()` in `src/modules/clustering/hierarchical_clustering_method.cpp`).

Implementation
----------------

We reuse an existing implementation of the hierarchical clustering algorithm from [this](https://github.com/cdalitz/hclust-cpp) repo which we add to the project as a submodule.

Review
-------

It is probably a good idea to review this PR on a commit-by-commit basis because this makes it easier to follow the changes.

EDIT (irallia, 13.04.2021): As this is the last PR of the epic, it resolves #32 aswell.


include/structures/cluster.hpp
```diff
@@ -43,6 +43,9 @@ class Cluster
 
     // \brief Returns the average length of the inserted sequences of all cluster members.
     int32_t get_average_inserted_sequence_size() const;
+
+    // \brief Returns the members of the cluster.
+    std::vector<Junction> get_members() const;
```
SGSSGene, line 48: Is your goal to have a compatibility to doxygen?

```suggestion
    //! \brief Returns the members of the cluster.
    std::vector<Junction> get_members() const;
```

include/structures/cluster.hpp
```diff
@@ -54,3 +57,12 @@ inline stream_t operator<<(stream_t && stream, Cluster const & clust)
...
 }
+
+/*! \brief A cluster is smaller than another, if its average first mate, average second mate, 
+ *         or average inserted sequence size (in this order) is smaller than the corresponding
+ *         element of the other cluster.
+ *
+ * \param lhs   left side cluster
+ * \param rhs   right side cluster
+ */
+bool operator<(Cluster const & lhs, Cluster const & rhs);
```
SGSSGene, line 68: There is missing a new line at the end of this file. This makes it technical not a text file.
Please add a newline at the end.
Which editor are you using?

lib/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+cmake_minimum_required (VERSION 3.8)
+
+add_library ("fastcluster" STATIC hclust/fastcluster.cpp)
+target_include_directories ("fastcluster" PUBLIC hclust)
```
SGSSGene, line 4: new line at the end of the file is missing.

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Partition based on mate 1
+    std::vector<Junction> current_partition{};
+    std::vector<std::vector<Junction>> current_partition_splitted;
+    std::vector<std::vector<Junction>> final_partitions{};
+
+    for (Junction junction : junctions)
+    {
+        if (current_partition.empty())
+        {
+            current_partition.push_back(junction);
```
SGSSGene, line 19: For maximum speed:
```suggestion
            current_partition.emplace_back(std::move(junction));
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+        else
+        {
+            if (junction.get_mate1().seq_name != current_partition.back().get_mate1().seq_name ||
+                junction.get_mate1().orientation != current_partition.back().get_mate1().orientation ||
+                abs(junction.get_mate1().position - current_partition.back().get_mate1().position) > 50)
+            {
+                // Partition based on mate 2
+                std::sort(current_partition.begin(), current_partition.end(), [](Junction a, Junction b) {
+                    return a.get_mate2() < b.get_mate2();
+                });
```
SGSSGene, line 30: for maximum speed:
```suggestion
                std::sort(current_partition.begin(), current_partition.end(), [](Junction const & a, Junction const & b) {
                    return a.get_mate2() < b.get_mate2();
                });
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+    {
+        std::sort(current_partition.begin(), current_partition.end(), [](Junction a, Junction b) {
+            return a.get_mate2() < b.get_mate2();
+        });
+        current_partition_splitted = split_partition_based_on_mate2(current_partition);
+        for (std::vector<Junction> partition : current_partition_splitted)
+        {
+            final_partitions.push_back(partition);
+        }
+    }
```
SGSSGene, line 51: Same as above.

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+        (lhs.get_mate2().seq_name == rhs.get_mate2().seq_name) &&
+        (lhs.get_mate2().orientation == rhs.get_mate2().orientation))
+    {
+        return (std::abs(lhs.get_mate1().position - rhs.get_mate1().position) +
+                std::abs(lhs.get_mate2().position - rhs.get_mate2().position) +
+                std::abs((int)(lhs.get_inserted_sequence().size() - rhs.get_inserted_sequence().size())));
+    }
+    else
+    {
+        return std::numeric_limits<int>::infinity();
```
SGSSGene, line 102: This doesn't do what anyone in the world would expect: https://en.cppreference.com/w/cpp/types/numeric_limits/infinity
It returns 0.... (What a stupid api).

```suggestion
        return std::numeric_limits<int>::max();
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+    for (std::vector<Junction> partition : partitions)
+    {
+        size_t partition_size = partition.size();
+        if (partition_size < 2)
+        {
+            clusters.emplace_back(partition);
+            continue;
+        }
+        // Compute condensed distance matrix (upper triangle of the full distance matrix)
+        double* distmat = new double[(partition_size * (partition_size - 1)) / 2];
```
SGSSGene, line 120: better don't use new/delete. Use a vector instead:

```suggestion
        std::vector<double> distmat((partition_size * (partition_size - 1)) / 2);
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+                distmat[k] = junction_distance(partition[i], partition[j]);
+                ++k;
+            }
+        }
+
+        // Perform hierarchical clustering
+        // `height` is filled with cluster distance for each step
+        // `merge` contains dendrogram
+        int* merge = new int[2 * (partition_size - 1)];
+        double* height = new double[partition_size - 1];
```
SGSSGene, line 134:
```suggestion

        std::vector<int> merge(2 * (partition_size - 1));
        std::vector<double> height(partition_size - 1);
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+        hclust_fast(partition_size, distmat, HCLUST_METHOD_AVERAGE, merge, height);
+
+        // Fill labels[i] with cluster label of junction i.
+        // Clustering is stopped at step with cluster distance >= clustering_cutoff
+        int* labels = new int[partition_size];
```
SGSSGene, line 139:
```suggestion
        std::vector<int> labels(partition_size);
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+        {
+            std::sort(jun.begin(), jun.end());
+            clusters.emplace_back(jun);
+        }
+
+        // Free memory
+        delete[] distmat;
+        delete[] merge;
+        delete[] height;
+        delete[] labels;
```
SGSSGene, line 165: If you use managed memory (in this case `vector`) no need to custom delete space.
```suggestion
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+        // Compute condensed distance matrix (upper triangle of the full distance matrix)
+        double* distmat = new double[(partition_size * (partition_size - 1)) / 2];
+        int k, i, j;
+        for (i = k = 0; i < partition_size; ++i) {
+            for (j = i + 1; j< partition_size; ++j) {
+                // Compute distance between junctions i and j  
+                distmat[k] = junction_distance(partition[i], partition[j]);
+                ++k;
+            }
+        }
```
SGSSGene, line 128: - Better prevent some leakage by defining variables inside the for loop.
- More consistency by using `size_t`.
- You can have multiple operation by seperating them by a comma: `++j, ++k`
```suggestion
        for (size_t i{0}, k{0}; i < partition_size; ++i) {
            for (size_t j{i + 1}; j< partition_size; ++j, ++k) {
                // Compute distance between junctions i and j  
                distmat[k] = junction_distance(partition[i], partition[j]);
            }
        }
```

src/structures/cluster.cpp
```diff
@@ -80,4 +80,22 @@ int32_t Cluster::get_average_inserted_sequence_size() const
...
+    return lhs.get_average_mate1() < rhs.get_average_mate1()
+            ? true
+            : rhs.get_average_mate1() < lhs.get_average_mate1()
+                ? false
+                : lhs.get_average_mate2() < rhs.get_average_mate2()
+                    ? true
+                    : rhs.get_average_mate2() < lhs.get_average_mate2()
+                        ? false
+                        : lhs.get_average_inserted_sequence_size() < rhs.get_average_inserted_sequence_size();
 }
```
SGSSGene, line 101: new line missing at end of file.

test/api/CMakeLists.txt
```diff
@@ -5,3 +5,5 @@ target_use_datasources (junction_detection_test FILES simulated.minimap2.hg19.co
 
 add_api_test (junction_detection_methods_test.cpp)
 target_use_datasources (junction_detection_methods_test FILES simulated.minimap2.hg19.coordsorted_cutoff.sam)
+
+add_api_test (clustering_methods_test.cpp)
```
SGSSGene, line 9: new line at end of file

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,169 @@
...
+                                                    double clustering_cutoff)
+{
+    auto partitions = partition_junctions(junctions);
+    std::vector<Cluster> clusters{};
+    for (std::vector<Junction> partition : partitions)
+    {
+        size_t partition_size = partition.size();
+        if (partition_size < 2)
+        {
+            clusters.emplace_back(partition);
```
SGSSGene, line 116: more speed!

```suggestion
            clusters.emplace_back(std::move(partition));
```

test/api/clustering_methods_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+const std::string chrom2 = "chr2";
+const int32_t chrom2_position1 = 234432;
+const std::string & read_name_1 = "m2257/8161/CCS";
+const std::string & read_name_2 = "m41327/11677/CCS";
+const std::string & read_name_3 = "m21263/13017/CCS";
+const std::string & read_name_4 = "m38637/7161/CCS";
+const std::string & read_name_5 = "m23412/9534/CCS";
+const std::string & read_name_6 = "m1245/5634/CCS";
+const std::string & read_name_7 = "m8765/9765/CCS";
+const std::string & read_name_8 = "m13456/11102/CCS";
```
SGSSGene, line 23: The references are pointing to temporary objects, and I believe are invalid after startup.
```suggestion
const std::string read_name_1 = "m2257/8161/CCS";
const std::string read_name_2 = "m41327/11677/CCS";
const std::string read_name_3 = "m21263/13017/CCS";
const std::string read_name_4 = "m38637/7161/CCS";
const std::string read_name_5 = "m23412/9534/CCS";
const std::string read_name_6 = "m1245/5634/CCS";
const std::string read_name_7 = "m8765/9765/CCS";
const std::string read_name_8 = "m13456/11102/CCS";
```

test/api/clustering_methods_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+                     Breakend{chrom2, chrom1_position3 - 1, strand::reverse}, ""_dna5, read_name_7}, //cluster 4
+            Junction{Breakend{chrom1, chrom1_position2 + 6, strand::forward},
+                     Breakend{chrom2, chrom1_position3 + 2, strand::reverse}, ""_dna5, read_name_8} //cluster 4
+        }
+    };
+
+    EXPECT_EQ(expected_partitions.size(), partitions.size());
+
+    if (expected_partitions.size() == partitions.size())
+    {
```
SGSSGene, line 84: Use `ASSERT_` instead of `EXPECT_EQ` to force gtest to abort this test. This way you can skip your if.
```suggestion
    ASSERT_EQ(expected_partitions.size(), partitions.size());
```

test/api/clustering_methods_test.cpp
```diff
@@ -0,0 +1,336 @@
...
+            if (expected_clusters[cluster_index].get_cluster_size() == clusters[cluster_index].get_cluster_size())
+            {
+                for (size_t junction_index = 0; junction_index < expected_clusters[cluster_index].get_cluster_size(); ++junction_index)
+                {
+                    EXPECT_TRUE(expected_clusters[cluster_index].get_members()[junction_index] == clusters[cluster_index].get_members()[junction_index]) << "Junction " << junction_index << " in cluster " << cluster_index << " different than expected";
+                }
+            }
+        }
+    }
+}
```
SGSSGene, line 336: please add a new line.

src/structures/cluster.cpp
```diff
@@ -80,4 +80,22 @@ int32_t Cluster::get_average_inserted_sequence_size() const
...
+{
+    return lhs.get_average_mate1() < rhs.get_average_mate1()
+            ? true
+            : rhs.get_average_mate1() < lhs.get_average_mate1()
+                ? false
+                : lhs.get_average_mate2() < rhs.get_average_mate2()
+                    ? true
+                    : rhs.get_average_mate2() < lhs.get_average_mate2()
+                        ? false
+                        : lhs.get_average_inserted_sequence_size() < rhs.get_average_inserted_sequence_size();
```
SGSSGene, line 100: Interesting... is this really the best we can do with c++? I just searched for it, but I couldn't find any better way...

SGSSGene (changes requested):
Changes look really great. :+1:

There are two hard errors:
1. `std::string const &  = "blub"` creates a reference onto a temporary reference...which makes any acces invalid (I believe, not sure why it didn't make any trouble).
2. `std::numeric_limits<int>::infinity()` should be replaced by `::max()`.
3. Cool c++ kids don't use new/delete :joy: 
The rest of my comments can definitely be ignored if you don't like :smiley: 

@File Ending: As far as I understand every line of a text file has to be finished by an '\n', otherwise it is technically not a text file and some programs will fail (I believe it is not even a valid c++ program, but no compiler complains). This actually should be cared for by your text editor automatically. Not sure what went wrong.

include/structures/cluster.hpp
eldariont: Well spotted! I'm using VS Code which has the option to insert a new line at the end of each saved file. Unfortunately, it was disabled for me, so I now turned it on.

src/modules/clustering/hierarchical_clustering_method.cpp
eldariont: Good point! Just to make sure I understood correctly: In each iteration of the for-loop, a new copy of the junction is created. If we simply `push_back()`, a second (unnecessary) copy is created.
Your suggestion now moves the first junction copy directly into the vector. Another similar option would be:
`current_partition.push_back(std::move(junction));`
This should also move the junction. Is that all correct or am I missing something?

eldariont: Another option that just came to my mind.. Would that have the same effect (only one copy per junction)?
```
for (Junction & junction : junctions)
{
    ... 
    current_partition.push_back(junction);
    ...
}
```

SGSSGene: Yes. This is even better! One small thing, the const goes to the right: `for (Junction const & junction : junctions)`

src/modules/clustering/hierarchical_clustering_method.cpp
eldariont: Good suggestion! I also changed the containing for-loop so that elements are accessed by non-const reference. This should avoid another copy operation, right?

test/api/clustering_methods_test.cpp
eldariont: Well, the references should extend the lifetime of the temporary objects so that they are accessible as long as the reference lives. But it would be simpler to just use values as you suggested.

test/api/clustering_methods_test.cpp
eldariont: Great suggestion! 👍

src/structures/cluster.cpp
eldariont: There is actually another but still similar way which is used in `breakend.cpp`. It first checks for inequality first and then checks which is smaller. Maybe that way is more understandable?!

```
    return lhs.seq_name != rhs.seq_name
            ? lhs.seq_name < rhs.seq_name
            : lhs.orientation != rhs.orientation
                ? lhs.orientation < rhs.orientation
                : lhs.position < rhs.position;
```

test/api/clustering_methods_test.cpp
SGSSGene: Yes, you are completely right! @Irallia I claimed something else in different PR, but that was wrong. 
It seems that a "const reference" extends the lifetime
https://herbsutter.com/2008/01/01/gotw-88-a-candidate-for-the-most-important-const/

eldariont: It's also in the C++ reference: https://en.cppreference.com/w/cpp/language/reference

*Rvalue references can be used to extend the lifetimes of temporary objects (note, lvalue references to const can extend the lifetimes of temporary objects too, but they are not modifiable through them)*

src/structures/cluster.cpp
SGSSGene: I like your suggestion. It is much clearer.

I think if you can use fields instead of function you can also do the following:
```
return std::tie(lhs.seq_name, lhs.orientation, lhs.position) < std::tie(rhs.seq_name, rhs.orientation, rhs.position);
```
But if you use function it might copy and be slower :-(

eldariont: I'm not sure I understand what you mean with using "fields instead of function"  (can you maybe explain a bit more?) but the expression you suggest looks much cleaner and shorter. According to the C++ reference, `std::tie` creates a tuple of references so it should be only a small overhead compared to using the const references in its present form.

eldariont:
Hi @SGSSGene, thanks for the very helpful review. I now incorporated most of your suggestions - with the exception of the "uncool" new/delete 😄 

The hierarchical clustering functions that I use (as a dependency) expect arrays/pointers as parameters.

From `lib/hclust/fastcluster.h`:
`int hclust_fast(int n, double* distmat, int method, int* merge, double* height);`

The size of the arrays is computed during runtime based on the number of elements in each partition. Currently, the space for the arrays is dynamically allocated using `new`. Do you see an option that would get rid of the new/delete? The `std::array` constructor expects a constexpr as size and `std::vector` does not translate into a pointer AFAIK. So which other options are there?

src/modules/clustering/hierarchical_clustering_method.cpp
SGSSGene: True, should work! Nice :-)

SGSSGene:
> Hi @SGSSGene, thanks for the very helpful review. I now incorporated most of your suggestions - with the exception of the "uncool" new/delete smile
> 
> The hierarchical clustering functions that I use (as a dependency) expect arrays/pointers as parameters.
> 
> From `lib/hclust/fastcluster.h`:
> `int hclust_fast(int n, double* distmat, int method, int* merge, double* height);`
> 
> The size of the arrays is computed during runtime based on the number of elements in each partition. Currently, the space for the arrays is dynamically allocated using `new`. Do you see an option that would get rid of the new/delete? The `std::array` constructor expects a constexpr as size and `std::vector` does not translate into a pointer AFAIK. So which other options are there?

Ah yes, I see. Checkout `std::vector::data()` https://en.cppreference.com/w/cpp/container/vector/data

You can use it as follows:
```
std::vector<int> myVector(10); // initializes the vector size 10
assert(myVector.size() == 10);
int* ptr = myVector.data();
```

At the end `std::vector` of course also calls new/delete, but it is much safer, since it handles many cases for exceptions and early returns.  The general concept is called RAII and can be extended from memory to many other resources https://en.cppreference.com/w/cpp/language/raii

src/structures/cluster.cpp
SGSSGene: > I'm not sure I understand what you mean with using "fields instead of function" (can you maybe explain a bit more?) but the expression you suggest looks much cleaner and shorter. According to the C++ reference, `std::tie` creates a tuple of references so it should be only a small overhead compared to using the const references in its present form.

Yeah, I wasn't very precise.
If you have for example `return std::tie(lhs.func1(), lhs.func2()) < std::tie(rhs.func1(), rhs.func2())`  All 4 functions have to be evaluated before the `<`-operator is being executed. If  `func1` and `func2` are very slow functions this might not be what you want and you want to use your original solution, which only executes `func1` and only if they are the same executes `func2`.
If you use variables (I called them accidentally fields) instead of functions, you don't have to worry about slow performing functions. In general this is all microoptimization. I got the feeling I am derailing your PR :disappointed: .
Any of your suggestions are fine. :smiley:

eldariont: Thanks for the explanation. Yes, that makes a lot of sense and I think it's always good to keep these smaller optimizations in mind 👍  So no worries, I'm very thankful for this kind of discussion 😃 

I now reimplemented the comparison operator for the `breakend` class because it accesses only member variables. However, I left the operators of the `junction` and `cluster` classes because these use functions which create additional `breakend` copies.

eldariont:
> Ah yes, I see. Checkout `std::vector::data()` https://en.cppreference.com/w/cpp/container/vector/data
> 
> You can use it as follows:
> 
> ```
> std::vector<int> myVector(10); // initializes the vector size 10
> assert(myVector.size() == 10);
> int* ptr = myVector.data();
> ```
> 
> At the end `std::vector` of course also calls new/delete, but it is much safer, since it handles many cases for exceptions and early returns. The general concept is called RAII and can be extended from memory to many other resources https://en.cppreference.com/w/cpp/language/raii

Great, I replaced new/delete with nice and safe std::vectors now. Is the `assert(v.size() == n)` required after I default-initialized the vector with `std::vector<T> v(n)`?

SGSSGene:
> > Ah yes, I see. Checkout `std::vector::data()` https://en.cppreference.com/w/cpp/container/vector/data
> > You can use it as follows:
> > ```
> > std::vector<int> myVector(10); // initializes the vector size 10
> > assert(myVector.size() == 10);
> > int* ptr = myVector.data();
> > ```
> > 
> > 
> > At the end `std::vector` of course also calls new/delete, but it is much safer, since it handles many cases for exceptions and early returns. The general concept is called RAII and can be extended from memory to many other resources https://en.cppreference.com/w/cpp/language/raii
> 
> Great, I replaced new/delete with nice and safe std::vectors now. Is the `assert(v.size() == n)` required after I default-initialized the vector with `std::vector<T> v(n)`?

The `assert(v.size() == n)` is not required. I just put it there to "prove" that it is guaranteed that v.size() is always equal to n.

SGSSGene:
Could you rebase this branch (or merge the origin/master into it)? there seems to some conflict in src/variant_detection/variant_detection.cpp

eldariont:
> Could you rebase this branch (or merge the origin/master into it)? there seems to some conflict in src/variant_detection/variant_detection.cpp

Hi @SGSSGene, I rebased the branch so it's ready for another review :)

src/structures/breakend.cpp
```diff
@@ -9,11 +11,7 @@
  */
 bool operator<(Breakend const & lhs, Breakend const & rhs)
 {
-    return lhs.seq_name != rhs.seq_name
-            ? lhs.seq_name < rhs.seq_name
-            : lhs.orientation != rhs.orientation
-                ? lhs.orientation < rhs.orientation
-                : lhs.position < rhs.position;
+    return std::tie(lhs.seq_name, lhs.orientation, lhs.position) < std::tie(rhs.seq_name, rhs.orientation, rhs.position);
```
Irallia, line 14: Oh nice, you found a better solution! 🎉

src/structures/cluster.cpp
```diff
@@ -80,4 +80,18 @@ int32_t Cluster::get_average_inserted_sequence_size() const
...
+    return members;
+}
+
+bool operator<(Cluster const & lhs, Cluster const & rhs)
+{
+    return lhs.get_average_mate1() != rhs.get_average_mate1()
+        ? lhs.get_average_mate1() < rhs.get_average_mate1()
+        : lhs.get_average_mate2() != rhs.get_average_mate2()
+            ? lhs.get_average_mate2() < rhs.get_average_mate2()
+            : lhs.get_average_inserted_sequence_size() < rhs.get_average_inserted_sequence_size();
```
Irallia, line 96: Can you use this std:tie here aswell?

src/structures/junction.cpp
```diff
@@ -22,18 +22,19 @@ std::string Junction::get_read_name() const
...
-                : lhs.get_mate2() < rhs.get_mate2()
-                    ? true
-                    : rhs.get_mate2() < lhs.get_mate2()
-                        ? false
-                        : lhs.get_inserted_sequence() < rhs.get_inserted_sequence();
+    return lhs.get_mate1() != rhs.get_mate1()
+        ? lhs.get_mate1() < rhs.get_mate1()
+        : lhs.get_mate2() != rhs.get_mate2()
+            ? lhs.get_mate2() < rhs.get_mate2()
+            : lhs.get_inserted_sequence() < rhs.get_inserted_sequence();
```
Irallia, line 29: and here?

test/api/CMakeLists.txt
Irallia: still open?

Irallia (changes requested):
Thank you for this work!
I have started to look at some of the details of the changes. I will do a full review after the approved first review.
But thank you very much, it looks great! 🎉

src/structures/cluster.cpp
eldariont: Yes, I could but decided against it. The `std::tie` creates one tuple each for the lhs and rhs and thus calls all six methods (including the copy operations needed in there). The current comparison first calls two methods and only if their value is equal calls the next two. So we can decide between elegance and speed here..

eldariont:
Hi @SGSSGene, I rebased the PR another time. As soon as you are satisfied with the changes, @Irallia can do the second review.

SGSSGene (approved):
I think your call against `std::tie` is a good call. It would harm performance.
I see that there are still a few `for` loops that could use more tidy up, but I think
it is not so important,  it doesn't harm readability and too much premature optimizations might lead to waste of development time. So from my sight it gets a green light

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,163 @@
...
+        std::vector<int> merge (2 * (partition_size - 1));
+        std::vector<double> height (partition_size - 1);
+        hclust_fast(partition_size, distmat.data(), HCLUST_METHOD_AVERAGE, merge.data(), height.data());
+
+        // Fill labels[i] with cluster label of junction i.
+        // Clustering is stopped at step with cluster distance >= clustering_cutoff
+        std::vector<int> labels (partition_size);
+        cutree_cdist(partition_size, merge.data(), height.data(), clustering_cutoff, labels.data());
+
+        std::unordered_map<int, std::vector<Junction>> label_to_junctions {};
```
Irallia, line 142:
```suggestion
        std::unordered_map<int, std::vector<Junction>> label_to_junctions{};
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,163 @@
...
+            if (label_to_junctions.find(labels[i]) != label_to_junctions.end())
+            {
+                label_to_junctions[labels[i]].push_back(std::move(partition[i]));
+            }
+            else{
+                label_to_junctions.emplace(labels[i], std::vector{std::move(partition[i])});
+            }
+        }
+
+        // Add new clusters
```
Irallia, line 154:
```suggestion
        // Add new clusters: Junctions with the same lable belong to one cluster
```

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
+        Junction{Breakend{chrom1, chrom1_position1 + 5, strand::forward},
+                    Breakend{chrom2, chrom2_position1 - 1, strand::reverse}, ""_dna5, read_name_4}, //cluster 2
+        Junction{Breakend{chrom1, chrom1_position1 + 92, strand::forward},
+                    Breakend{chrom2, chrom2_position1 + 3, strand::forward}, ""_dna5, read_name_5}, //cluster 3
+        Junction{Breakend{chrom1, chrom1_position2 - 2, strand::forward},
+                    Breakend{chrom2, chrom1_position3 + 8, strand::reverse}, ""_dna5, read_name_6}, //cluster 4
+        Junction{Breakend{chrom1, chrom1_position2 + 3, strand::forward},
+                    Breakend{chrom2, chrom1_position3 - 1, strand::reverse}, ""_dna5, read_name_7}, //cluster 4
+        Junction{Breakend{chrom1, chrom1_position2 + 6, strand::forward},
+                    Breakend{chrom2, chrom1_position3 + 2, strand::reverse}, ""_dna5, read_name_8} //cluster 4
```
Irallia, line 57:
```suggestion
    std::vector<Junction> input_junctions
    {   
        Junction{Breakend{chrom1, chrom1_position1 - 5, strand::forward},
                 Breakend{chrom2, chrom2_position1 + 8, strand::forward}, ""_dna5, read_name_1}, //cluster 1
        Junction{Breakend{chrom1, chrom1_position1 + 2, strand::forward},
                 Breakend{chrom2, chrom2_position1 - 3, strand::forward}, ""_dna5, read_name_2}, //cluster 1
        Junction{Breakend{chrom1, chrom1_position1 + 9, strand::forward},
                 Breakend{chrom2, chrom2_position1 + 1, strand::forward}, ""_dna5, read_name_3}, //cluster 1
        Junction{Breakend{chrom1, chrom1_position1 + 5, strand::forward},
                 Breakend{chrom2, chrom2_position1 - 1, strand::reverse}, ""_dna5, read_name_4}, //cluster 2
        Junction{Breakend{chrom1, chrom1_position1 + 92, strand::forward},
                 Breakend{chrom2, chrom2_position1 + 3, strand::forward}, ""_dna5, read_name_5}, //cluster 3
        Junction{Breakend{chrom1, chrom1_position2 - 2, strand::forward},
                 Breakend{chrom2, chrom1_position3 + 8, strand::reverse}, ""_dna5, read_name_6}, //cluster 4
        Junction{Breakend{chrom1, chrom1_position2 + 3, strand::forward},
                 Breakend{chrom2, chrom1_position3 - 1, strand::reverse}, ""_dna5, read_name_7}, //cluster 4
        Junction{Breakend{chrom1, chrom1_position2 + 6, strand::forward},
                 Breakend{chrom2, chrom1_position3 + 2, strand::reverse}, ""_dna5, read_name_8} //cluster 4
```

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
+
+
+TEST(clustering, partitioning)
+{
+    std::vector<Junction> input_junctions = prepare_input_junctions();
+    std::vector<std::vector<Junction>> partitions = partition_junctions(input_junctions);
+
+    std::vector<std::vector<Junction>> expected_partitions
+    {
+        {   Junction{Breakend{chrom1, chrom1_position1 - 5, strand::forward},
```
Irallia, line 72: Since we are following seqan3 format claims:
```suggestion
        {   
            Junction{Breakend{chrom1, chrom1_position1 - 5, strand::forward},
```

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
+            Junction{Breakend{chrom1, chrom1_position1 + 9, strand::forward},
+                     Breakend{chrom2, chrom2_position1 + 1, strand::forward}, ""_dna5, read_name_3}, //cluster 1
+        },
+        {   Junction{Breakend{chrom1, chrom1_position1 + 5, strand::forward},
+                     Breakend{chrom2, chrom2_position1 - 1, strand::reverse}, ""_dna5, read_name_4}, //cluster 2
+        },
+        {   Junction{Breakend{chrom1, chrom1_position1 + 92, strand::forward},
+                     Breakend{chrom2, chrom2_position1 + 3, strand::forward}, ""_dna5, read_name_5}, //cluster 3
+        },
+        {   Junction{Breakend{chrom1, chrom1_position2 - 2, strand::forward},
```
Irallia, line 85: same here

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
+                             Breakend{chrom2, chrom1_position3 + 2, strand::reverse}, ""_dna5, read_name_8}
+        }}
+    };
+    std::sort(expected_clusters.begin(), expected_clusters.end());
+
+    ASSERT_EQ(expected_clusters.size(), clusters.size());
+
+    for (size_t cluster_index = 0; cluster_index < expected_clusters.size(); ++cluster_index)
+    {
+        ASSERT_EQ(expected_clusters[cluster_index].get_cluster_size(), clusters[cluster_index].get_cluster_size()) << "Cluster " << cluster_index << " of unexpected size";
```
Irallia, line 146: is it possible to split this line?
By the way, I don't know this `ASSERT(...) << bla` at all, that you can write an output behind it. You never stop learning. :)

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
+
+        for (size_t junction_index = 0; junction_index < expected_clusters[cluster_index].get_cluster_size(); ++junction_index)
+        {
+            EXPECT_TRUE(expected_clusters[cluster_index].get_members()[junction_index] == clusters[cluster_index].get_members()[junction_index]) << "Junction " << junction_index << " in cluster " << cluster_index << " different than expected";
```
Irallia, line 150: and here aswell?

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
```
Irallia, line 207: and here

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
```
Irallia, line 262: and here

test/api/clustering_test.cpp
```diff
@@ -1,19 +1,319 @@
...
```
Irallia, line 316: and the last of these

CMakeLists.txt
```diff
@@ -30,6 +30,9 @@ find_package (SeqAn3 QUIET REQUIRED HINTS lib/seqan3/build_system)
 add_subdirectory (src)
 message (STATUS "${FontBold}You can run `make` to build the application.${FontReset}")
 
+# Dependency: hclust.
+add_subdirectory (lib)
+
```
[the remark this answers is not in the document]
Irallia: no?

Irallia (changes requested):
Wow, that was a lot to read and process. 😮
First of all, thank you! Thanks also for the detailed tests on the new method! ❤️ 
I have a few documentation suggestions and otherwise really only style issues. 💅 
Content wise it looks great! I hope I understood everything  😄

include/modules/clustering/hierarchical_clustering_method.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include "structures/cluster.hpp"   // for class Cluster
+
+
+/*! \brief Partition junctions by their distance on the reference genome.
+ *         The junctions in each of the returned partitions are sorted even though
+ *         the partitions themselves are not returned in a particular order.
+ *
```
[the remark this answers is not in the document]
eldariont: Great suggestion!

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -0,0 +1,163 @@
...
+    }
+    if (!current_partition.empty())
+    {
+        std::sort(current_partition.begin(), current_partition.end());
+        splitted_partition.push_back(current_partition);
+    }
+    return splitted_partition;
+}
+
+int junction_distance(Junction const & lhs, Junction const & rhs)
```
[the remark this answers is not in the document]
eldariont: Well spotted. I added the declaration to the hpp file.

test/api/clustering_test.cpp
eldariont: https://github.com/google/googletest/blob/master/docs/primer.md#assertions

CMakeLists.txt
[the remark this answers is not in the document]
eldariont: Yes, it's much better to get rid of the git submodule because we just want to use the library instead of accessing the source code. I just pushed a commit that replaces the submodule with CMake's `FetchContent` to get the hierarchical clustering dependency. For `FetchContent` I needed to increase the CMake version to 3.11.

include/modules/clustering/hierarchical_clustering_method.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "structures/cluster.hpp"   // for class Cluster
+
+/*! \brief Partition junctions by their distance on the reference genome.
+ *         The returned partitions contain junctions meeting the following criteria:
+ *         a) all junctions in a partition connect the same reference sequences,
+ *         b) all junctions in a partition have the same orientations, and
+ *         c) the distance between corresponding mates of two neighboring junctions is at most 50bp.
+ *         The junctions in each of the returned partitions are sorted even though
+ *         the partitions themselves are not returned in a particular order.
```
Irallia, line 11:
```suggestion
/*! \brief Partition junctions by their distance on the reference genome.
 *         The returned partitions contain junctions meeting the following criteria:
 *         a) all junctions in a partition connect the same reference sequence,
 *         b) all junctions in a partition have the same orientation, and
 *         c) the distance between corresponding mates of two neighboring junctions is at most 50bp.
 *         The junctions in each of the returned partitions are sorted even though
 *         the partitions themselves are not returned in a particular order.
```

include/modules/clustering/hierarchical_clustering_method.hpp
```diff
@@ -0,0 +1,44 @@
...
+ * \param[in] partition - a partition (i.e. a vector) of junctions (needs to be sorted
+ *                        by their second mates)
+ */
+std::vector<std::vector<Junction>> split_partition_based_on_mate2(std::vector<Junction> const & partition);
+
+/*! \brief Compute the distance between two junctions.
+ *         For two junctions that connect the same referene sequences and have the same
+ *         orientations, the distance is the sum of a) the distance between the first mates,
+ *         b) the distance between the second mates, and c) the absolute size difference 
+ *         of the inserted sequences. For other junctions, the distance takes the maximal value.
```
Irallia, line 30:
```suggestion
 *         For two junctions that connect the same reference sequence and have the same
 *         orientation, the distance is the sum of a) the distance between the first mates,
 *         b) the distance between the second mates, and c) the absolute size difference 
 *         of the inserted sequences. For other junctions, the distance takes the maximal value.
```

Irallia (approved):
Thanks for the changes! LGFM now, just two misspellings.
I'll commit These last changes by myself and merge this PR afterwards.

Issue #87, opened 2021-03-15
[INFRA]  Restructure `detect_variants_in_alignment_file()`


Irallia:
Restructure huge `detect_variants_in_alignment_file()` into a workflow of functions, which can be called seperately.
Also change the passing parameters of this function, maybe pass the args structure or create a structure which could be passed.

A follow up would be, to create new tests and especially rewrite the methods tests.

```cpp
detect_variants_in_alignment_file(...)
{
    detect();
    cluster();
    refine();
    ...
}
```
and move this function/functionality into `iGenVar.cpp`?

PR #103, opened 2021-04-12
[MISC] pass args instead of individual arguments

Irallia:
Resolves first part of https://github.com/seqan/iGenVar/issues/87 

This PR is mostly moving functions from one file into another.

### DIFF
--- a/include/variant_detection/method_enums.hpp
+++ b/include/variant_detection/method_enums.hpp
@@ -1,5 +1,8 @@
 #pragma once
 
+#include <string>           // for std::string
+#include <unordered_map>    // for std::unordered_map
+
 //!\brief An enum for the different clustering methods.
 enum detection_methods
 {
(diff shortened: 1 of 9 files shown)


MitraDarja (approved):

include/variant_detection/method_enums.hpp
```diff
@@ -32,3 +35,21 @@ enum refinement_methods
     sViper_refinement_method = 1,
     sVirl_refinement_method = 2
 };
+
+/*! \brief Specialise a mapping from an identifying string to the respective value of your type methods. With the help
```
eldariont, line 39:
```suggestion
/*! \brief Specialise a mapping from an identifying string to the respective value of your type detection_methods. With the help
```

src/iGenVar.cpp
```diff
@@ -1,72 +1,9 @@
...
-{
-    std::filesystem::path alignment_short_reads_file_path{""};
-    std::filesystem::path alignment_long_reads_file_path{""};
-    std::filesystem::path insertion_file_path{};
-    std::filesystem::path output_file_path{};
-    std::vector<detection_methods> methods{cigar_string, split_read, read_pairs, read_depth};   // default: all methods
-    clustering_methods clustering_method{simple_clustering};                                    // default: simple clustering method
-    refinement_methods refinement_method{no_refinement};                                        // default: no refinement
-    uint64_t min_var_length = 30;
-};
```
eldariont, line 68: Why do you move the `cmd_arguments` struct to `variant_detection.hpp`? Wasn't it quite logical to have it close to the argument parsing method?

eldariont (approved):
Looks great. Only two smaller things..

src/iGenVar.cpp
Irallia: I will move it back in the next PR (#104) it was necessary as `cmd_arguments` is not known in `variant_detection.hpp`, but in the next PR I'll move the whole function into iGenVar.cpp

Irallia: I added a TODO for this.

Issue #12, opened 2020-09-24
[TEST] Write API tests for functions in iGenVar


Irallia:
src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
Write tests for:
- [x] `split_string()`
- [x] `retrieve_aligned_segments()`
- [x] `analyze_aligned_segments()`
- [x] `analyze_sa_tag()`

src/modules/sv_detection_methods/analyze_cigar_method.cpp
Write tests for:
- [x] `analyze_cigar()`

src/modules/clustering/simple_clustering_method.cpp
Write tests for:
- [x] `simple_clustering_method()`

PR #104, opened 2021-04-12
[TEST, INFRA] move functions and create tests

Irallia:
Follow up of https://github.com/seqan/iGenVar/pull/103
Resolves https://github.com/seqan/iGenVar/issues/87 (second part)

This PR also adds some tests for the api (https://github.com/seqan/iGenVar/issues/12)

I moved `detect_variants_in_alignment_file()` into the `iGenVar.cpp`
I also created an `iGenVar.hpp`
I also created a new order/system for the tests, we now have a file for each method type (detection, clustering, refinement) and for the input aswell (short and long read files)

EDIT (13.04.2021): The new tests are more or less place holders and need better mocked data.


Irallia:
The new function description looks like that:
Bildschirmfoto 2021-04-12 um 17 27 35

include/iGenVar.hpp
```diff
@@ -0,0 +1,39 @@
...
+void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments & args);
+
+/*! \brief Detects genomic variants by analyzing an alignment file (sam/bam). The detected
+ *         variants are printed to a given file or stdout and insertion alleles are stored in a fasta file.
+ *
+ * \param[in]   args - command line arguments:\n
+ *                     **args.alignment_short_reads_file_path** - short reads input file, path to the sam/bam file\n
+ *                     **args.alignment_long_reads_file_path** - long reads input file, path to the sam/bam file\n
+ *                     **args.output_file_path** output file - path for the VCF file
+ *                     **args.methods** - list of methods for detecting junctions (1: cigar_string,
```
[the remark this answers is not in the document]
joshuak94: Or just have (default) at the end of the option.

src/iGenVar.cpp
```diff
@@ -55,6 +69,69 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
...
+            break;
+        case 1: // sViper_refinement_method
+            seqan3::debug_stream << "The sViper refinement method is not yet implemented\n";
+            break;
+        case 2: // sVirl_refinement_method
+            seqan3::debug_stream << "The sVirl refinement method is not yet implemented\n";
+            break;
+    }
+
+    find_and_output_variant(clusters, args.output_file_path);
```
joshuak94, line 132: Super small optional thing: I would love this function to be `find_and_output_variants` (the "s" at the end lol)...

joshuak94 (approved):
Just a few small things!

test/api/detection_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+// TODO (irallia): implement test cases
+
+TEST(junction_detection, cigar_string_method_simple)
+{
+    std::string const read_name = "read021";
+    std::string const chromosome = "chr1";
+    int32_t const query_start_pos = 1;
+    std::vector<seqan3::cigar> cigar_sting = {{5, 'S'_cigar_operation}, {15, 'M'_cigar_operation},
+                                              {6, 'D'_cigar_operation}, {24, 'M'_cigar_operation}}; //5S15M6D24M
+    seqan3::dna5_vector seq = {"GGGCTCATCGATCGATTTCGGATCGGGGGGCCCCCATTTTAAACGGCCCC"_dna5};
```
eldariont, line 29: The CIGAR string indicates a sequence size of 5+15+24=44 but the sequence itself is 50bp long. The simplest solution would be to extend the final match operation from 24 to 30.

test/api/detection_test.cpp
```diff
@@ -0,0 +1,94 @@
...
+TEST(junction_detection, split_read_method_simple)
+{
+    std::string const read_name = "read021";
+    seqan3::sam_flag const flag{0u};
+    std::string const chromosome = "chr1";
+    int32_t const pos = 80;
+    uint8_t const mapq = 60;
+    std::vector<seqan3::cigar> cigar_sting = {{46, 'M'_cigar_operation}, {4, 'S'_cigar_operation}}; //46M4S
+    seqan3::dna5_vector seq = {"GGGCTCATCGATCGATTTCGGATCGGGGGGCCCCCATTTTAAACGGCCCC"_dna5};
+    std::string const sa_tag = "SA:Z:chr1,108,+,44S6M,60,6,0;";
```
eldariont, line 56: The two CIGAR strings indicate a slight 2bp overlap on the read (i.e. the 45. and 46. read base are aligned twice). It is checked [here](https://github.com/seqan/iGenVar/blob/f92d9c24de496d35a24721b9b031ebb50323f601/src/modules/sv_detection_methods/analyze_sa_tag_method.cpp#L65) that no such overlap exists before detecting a junction. So if you remove the overlap, `junction_res` should contain a junction which you can test for in `junctions_expected_res`.

test/api/detection_test.cpp
```diff
@@ -0,0 +1,94 @@
+#include <fstream>
+#include <gtest/gtest.h>
+
+#include <seqan3/alphabet/cigar/cigar.hpp>
+#include <seqan3/alphabet/nucleotide/dna5.hpp>
+#include <seqan3/io/exception.hpp>
+#include <seqan3/io/sam_file/sam_flag.hpp>
+
+#include "iGenVar.hpp"
```
eldariont, line 9: Why do you include `iGenVar.hpp` here?

eldariont (changes requested):
Thanks for this big chunk of work! This PR greatly improves the structure of our code and particularly the tests. I love how the tests are now nicely separated into detection, clustering and refinement tests 👍 

I had quite a few comments but they are mostly small things.

include/iGenVar.hpp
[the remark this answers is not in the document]
Irallia: How about this solution?
Bildschirmfoto 2021-04-14 um 11 51 52

test/api/detection_test.cpp
Irallia: Thanks for the hint. I found another mistake in our mini_example: https://github.com/seqan/iGenVar/issues/108

src/iGenVar.cpp
```diff
@@ -3,8 +3,10 @@
 #include <seqan3/core/debug_stream.hpp>                     // for seqan3::debug_stream
 
 #include "modules/clustering/simple_clustering_method.hpp"  // for the simple clustering method
-#include "variant_detection/validator.hpp"                  // for class EnumValidator
 #include "structures/cluster.hpp"                           // for class Cluster
+#include "variant_detection/method_enums.hpp"               // for enum detection_, clustering_ and refinement_methods
```
eldariont, line 7: I think this include is not strictly needed because it's already included in `variant_detection.hpp` but it probably doesn't hurt and is more verbose this way.

eldariont (approved):
Perfect 👍

PR #106, opened 2021-04-12
[FEATURE] handle short reads add empty read pair method

Irallia:
This resolves the first part of https://github.com/seqan/iGenVar/issues/105

Here I add an empty read pair method and handle short reads next to long reads.

### DIFF
--- a/include/variant_detection/bam_functions.hpp
+++ b/include/variant_detection/bam_functions.hpp
@@ -5,6 +5,28 @@
 #include <seqan3/io/sam_file/output.hpp>
 #include <seqan3/utility/char_operations/predicate.hpp>
 
+
+/*!\details
+ * From the SAM/BAM Format Specification
+ * | Bit  | Bit   | Description                                                        | Note                                               |
+ * |:----:|:-----:|:------------------------------------------------------------------:|:--------------------------------------------------:|
+ * |    1 | 0x1   | template having multiple segments in sequencing                    | necessary for read pair method                     |
+ * |    2 | 0x2   | each segment properly aligned according to the aligner             | aligned as expected, probably not containing SVs *)|
+ * |    4 | 0x4   | segment unmapped                                                   | will be skipped                                    |
+ * |    8 | 0x8   | next segment in the template unmapped                              | mate unmapped, maybe an insertion? *)              |
+ * |   16 | 0x10  | SEQ being reverse complemented                                     | important for split reads and paired-reads         |
+ * |   32 | 0x20  | SEQ of the next segment in the template being reverse complemented | *)                                                 |
+ * |   64 | 0x40  | the first segment in the template                                  | *)                                                 |
+ * |  128 | 0x80  | the last segment in the template                                   | *)                                                 |
+ * |  256 | 0x100 | secondary alignment                                                | will be skipped                                    |
+ * |  512 | 0x200 | not passing filters, such as platform/vendor quality controls      | *)                                                 |
+ * | 1024 | 0x400 | PCR or optical duplicate                                           | will be skipped                                    |
+ * | 2048 | 0x800 | supplementary alignment                                            | will be skipped (primary alignments only)          |
+ * *) not checked yet
+ * \see [Map Format Specification](https://samtools.github.io/hts-specs/SAMv1.pdf#page=7)
+ * (last access 09.04.2021).
+ *
+ */
 enum BamFlags
 {
     BAM_FLAG_MULTIPLE      = 0x0001,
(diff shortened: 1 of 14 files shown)


Irallia:
I also added some documentation e.g. this table:
Bildschirmfoto 2021-04-09 um 16 37 03

include/variant_detection/bam_functions.hpp
```diff
@@ -5,6 +5,26 @@
...
+
+/*!\details
+ * From the SAM/BAM Format Specification
+ * | Bit  | Bit   | Description                                                       | Note                           |
+ * |:----:|:-----:|:-----------------------------------------------------------------:|:------------------------------:|
+ * |    1 | 0x1   | template having multiple segments in sequencing                   |                                |
+ * |    2 | 0x2   | each segment properly aligned according to the aligner            | necessary for read pair method |
+ * |    4 | 0x4   | segment unmapped                                                  | will be skipped                |
+ * |    8 | 0x8   | next segment in the template unmapped                             |                                |
+ * |   16 | 0x10  | SEQbeing reverse complemented                                     | is checked in split read method |
```
simonsasse, line 17: Is there a blank space missing in this and in the next line ?
SEQ beging ..
SEQ of ...

src/variant_detection/variant_detection.cpp
```diff
@@ -4,28 +4,94 @@
...
+    for (auto & record : alignment_short_reads_file)
+    {
+        seqan3::sam_flag const flag         = record.flag();                            // 2: FLAG
+        int32_t const ref_id                = record.reference_id().value_or(-1);       // 3: RNAME
+        int32_t const ref_pos               = record.reference_position().value_or(-1); // 4: POS
+        uint8_t const mapq                  = record.mapping_quality();                 // 5: MAPQ
+        auto const header_ptr               = record.header_ptr();
+        auto const ref_ids = header_ptr->ref_ids();
+
+        // TODO (irallia): check for mapping quality (e.g. vaquita: see minMapQual)
```
simonsasse, line 43: Is this TODO outdated, because you check for `mapq < 20` in the next if clause ?

src/variant_detection/variant_detection.cpp
```diff
@@ -75,22 +141,18 @@ void detect_junctions_in_long_reads_sam_file(std::vector<Junction> & junctions,
...
                     break;
-                    // continue;
             }
         }
 
         num_good++;
         if (num_good % 1000 == 0)
         {
-            seqan3::debug_stream << num_good << " good alignments" << std::endl;
+            seqan3::debug_stream << num_good << " good alignments from long read file" << std::endl;
```
simonsasse, line 155:
```suggestion
            seqan3::debug_stream << num_good << " good alignments from long read file." << std::endl;
```

src/variant_detection/variant_detection.cpp
```diff
@@ -4,28 +4,94 @@
...
+                case detection_methods::read_depth: // Detect junctions from read depth evidence
+                    seqan3::debug_stream << "The read depth method short reads  is not yet implemented.\n";
+                    break;
+            }
+        }
+
+        num_good++;
+        if (num_good % 1000 == 0)
+        {
+            seqan3::debug_stream << num_good << " good alignments from short read file" << std::endl;
```
simonsasse, line 74:
```suggestion
            seqan3::debug_stream << num_good << " good alignments from short read file." << std::endl;
```

simonsasse (changes requested):
Voila :)

simonsasse (approved):
Looks good 💯

src/variant_detection/variant_detection.cpp
```diff
@@ -4,28 +4,92 @@
...
+                    seqan3::debug_stream << "The split cigar string method for short reads is not yet implemented.\n";
+                    break;
+                case detection_methods::split_read:     // Detect junctions from split read evidence (SA tag,
+                    seqan3::debug_stream << "The split read method for short reads is not yet implemented.\n";
+                    break;
+                case detection_methods::read_pairs: // Detect junctions from read pair evidence
+                    if (hasFlagProperMappedPairedReads(flag))
+                    {
+                        analyze_read_pair();
+                    }
```
eldariont, line 61: The `0x2` (properly mapped pair) flag is not well defined by the SAM specification but it usually means that the two mates are mapped in the right orientation and distance to each other. Analyzing only these reads, however, would be very boring and would ignore all the interesting reads covering SVs (which is what we actually want to find). So I would rather look at flag `0x1` (read paired).

include/variant_detection/bam_functions.hpp
```diff
@@ -5,6 +5,26 @@
 #include <seqan3/io/sam_file/output.hpp>
 #include <seqan3/utility/char_operations/predicate.hpp>
 
+
+/*!\details
+ * From the SAM/BAM Format Specification
+ * | Bit  | Bit   | Description                                                        | Note                                      |
+ * |:----:|:-----:|:------------------------------------------------------------------:|:-----------------------------------------:|
+ * |    1 | 0x1   | template having multiple segments in sequencing                    |                                           |
+ * |    2 | 0x2   | each segment properly aligned according to the aligner             | necessary for read pair method            |
```
eldariont, line 14: I do not agree with some of the notes. Here are a few suggestions:
- `0x1` - necessary for read pair method
- `0x2` - aligned as expected, probably not containing SVs
- `0x8` - mate unmapped, maybe an insertion?
- Flag `0x10` is not only important for split reads but also for paired-reads where `0x10` and `0x20` are compared. Usually you would expect them to be different and if they are not it could be because of an inversion. I don't know how to put this into a good concise note so you could also leave it empty.

include/variant_parser/variant_record.hpp
```diff
@@ -6,10 +6,12 @@
 
 #include <seqan3/io/stream/concept.hpp>
 
-/*
+/*!\details
  * An info field looks as follows:
- * ##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version">
- * https://samtools.github.io/hts-specs/VCFv4.3.pdf (1.4.2 Information field format)
+ *
+ * `##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version"`
```
eldariont, line 12:
```suggestion
 * `##INFO=<ID=ID,Number=number,Type=type,Description="description",Source="source",Version="version">`
```

test/api/input_file_test.cpp
```diff
@@ -51,15 +51,40 @@ void check_output(std::string const expected_res)
     return;
 }
 
-std::vector<Junction> junctions_res{};
+TEST(junction_detection, detect_junctions_in_short_read_sam_file)
```
eldariont, line 54:
```suggestion
TEST(input_file, detect_junctions_in_short_read_sam_file)
```

test/api/input_file_test.cpp
```diff
@@ -51,15 +51,40 @@ void check_output(std::string const expected_res)
...
+        //                      << (junctions_expected_res[i].get_mate1() == junctions_res[i].get_mate1()) << ": \n"
+        //                      << junctions_expected_res[i].get_mate1() << " == " << junctions_res[i].get_mate1() << "\n"
+        //                      << (junctions_expected_res[i].get_mate2() == junctions_res[i].get_mate2()) << ": \n"
+        //                      << junctions_expected_res[i].get_mate2() << " == " << junctions_res[i].get_mate2() << "\n";
+    }
+}
 
-TEST(junction_detection, detect_junctions_in_long_reads_sam_file)
+
+TEST(junction_detection, detect_junctions_in_long_read_sam_file)
```
eldariont, line 81:
```suggestion
TEST(input_file, detect_junctions_in_long_read_sam_file)
```

test/api/input_file_test.cpp
```diff
@@ -143,6 +168,7 @@ TEST(junction_detection, detect_junctions_in_long_reads_sam_file)
 
 TEST(junction_detection, long_read_sam_file_unsorted)
```
eldariont, line 169:
```suggestion
TEST(input_file, long_read_sam_file_unsorted)
```

eldariont (changes requested):
Thanks for the nice work! :+1: Just a few small things :)

src/variant_detection/variant_detection.cpp
Irallia: aaaah that explains why I was so confused. Thank you!

include/variant_detection/bam_functions.hpp
```diff
@@ -5,6 +5,26 @@
 #include <seqan3/io/sam_file/output.hpp>
 #include <seqan3/utility/char_operations/predicate.hpp>
 
+
+/*!\details
+ * From the SAM/BAM Format Specification
+ * | Bit  | Bit   | Description                                                        | Note                                       |
+ * |:----:|:-----:|:------------------------------------------------------------------:|:------------------------------------------:|
+ * |    1 | 0x1   | template having multiple segments in sequencing                    | necessary for read pair method             |
+ * |    2 | 0x2   | each segment properly aligned according to the aligner             | not checked yet - aligned as expected, probably not containing SVs |
```
eldariont, line 14: Does the `not checked yet` mean that this flag is not yet evaluated by iGenVar? If yes, `0x20`, `0x40`,`0x80` and `0x200` are also not checked yet. To save some space you could add a star * to every flag that is not evaluated yet and explain the star below the table?

eldariont (approved):
Looks very good. Thanks for your work 👍 
Just one little idea that you could follow or not.

Irallia:
Bildschirmfoto 2021-05-03 um 17 28 19

PR #107, opened 2021-04-13
[TEST] add paired end example

### LINKED ISSUE #101: [MISC] Add paired end reads to mini example

Irallia:
Resolves #101


eldariont (changes requested):
Thanks for working on this example! 🙂 During the review I discovered a few misconceptions I had about paired-end alignments:
- As it turns out, the read names of the two mates have to be identical (Quoting from the SAM specs: `QNAME: Query template NAME. Reads/segments having identical QNAME are regarded to come from the same template.`)
- The TLEN needs to be positive for the first and negative for the second mate (quoting again from the specs: `if all segments are mapped to the same reference sequence, the  absolute value of TLEN equals the distance between the mapped end of the template  and  the mapped start of the template, inclusively (i.e., end−start + 1). The TLEN field is positive for the leftmost segment of the template, negative for the rightmost, and the sign for any middle segment is undefined.`)

I also saw that the query sequences of your second mates are wrong. The query sequence (SEQ field) is always in forward orientation. If the mate is mapped in reverse orientation, the reverse complement of its sequence is used as SEQ. Quoting from the specs: `FLAG's bit 0x10 indicates whether SEQ has been reverse complemented and QUAL reversed. When bit 0x4 is unset, this corresponds to the strand to which the segment has been mapped: bit 0x10 unset indicates the forward strand, while set indicates the reverse strand.`

When making all these changes to `read001` I get the following lines (**EDIT: Corrected a few mistakes**):
```
read001	99	chr1	1	60	50M	chr1	113	148	CGCCCATGCAACTAGCGATGCTAGCTAGCTAGCTTACGACTGGCCATGCG	*	AS:i:50	NM:i:0
read001	147	chr1	114	60	12M15I23M	chr1	1	-148	CATTTTAAACGGCCCCGGGGCCAATTTGGATCTTGACTCTGGAAAACTTT	*	AS:i:31	NM:i:15
```

After making the changes for all reads, IGV should show no colors on your reads (colors represent mismatching bases). If you right-click on the file name and select `View as pairs`, the two mates of a read should be connected by a line.

Very sorry for all the confusion and the misconceptions about paired reads that I shared with you. I hope it's not too much work to make the changes now 🙈

Irallia:
paired_end_IGV_View

eldariont (approved):
Thanks for fixing all these different problems! All the numbers are correct now. And it's rather reassuring that IGV shows this clean display now. 👍

README.md
```diff
@@ -18,5 +18,6 @@ Instructions:
 
 (Built using the [SeqAn3 App Template](https://github.com/seqan/app-template))
 
-We created a small example, which you can use to test run our app:
-`./bin/iGenVar -j ./test/data/mini_example.sam -o ./test/data/output.vcf -m cigar_string -m split_read -l 5`
+We created small examples, which you can use to test run our app:
```
MitraDarja, line 21: Maybe just to "to test our app"? Idk, but to test run sounds weird to me.

MitraDarja (approved):
LGTM, just a minor thing.

PR #109, opened 2021-04-14
[FIX] Fix SA tags in mini_example

### LINKED ISSUE #108: [BUG] SA string in mini_example too long
Our SA strings have 7 fields, but there are only 6 exisiting. The `AS` is not a part of the SA string.

This is testet in `retrieve_aligned_segments()`, where the SA string is only handled, if its exactly of size 6
```if (fields.size() == 6)```
We could maybe add a warning, if a wrong sized SA string is existent.

### PR DESCRIPTION
**Pull request by Irallia**
Resolves #108

### CODE REVIEW

File: src/modules/sv_detection_methods/analyze_sa_tag_method.cpp [hunk 1]
```diff
@@ -47,6 +47,10 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
             int32_t mapq = std::stoi(fields[4]);
             aligned_segments.push_back(AlignedSegment{orientation, ref_name, pos, mapq, cigar_vector});
         }
+        else
+        {
+            seqan3::debug_stream << "Your SA tag has a wrong format (wrong amount of parameters): " << sa_tag << '\n';
```
joshuak94, line 52: Can we just have here a throw `std::runtime_error`?

joshuak94: Also is it even worth having this like this? The SA tag will be generated by alignment tools right, not by hand. So the odds of there being an error there would be basically none... So maybe it's more efficient to just take the if statement out.

Irallia: Hmmm, that is a good question. What do you think @eldariont ? I'm really not sure.

eldariont: Unfortunately, the odds of having a strange number of fields are not none. Last year, I was contacted by a user who reported that SVIM crashed on his data. When investigating the issue, I found that his chromosome names looked like this: `>Scaffold1;SomeOtherInfo=1`. The semi-colon in there caused the SA tags to be splitted into more than six fields. Therefore, I added the check to make sure that an error is raised as close to the problem as possible.
I don't mind whether we throw `std::runtime_error` or issue a debug message. The second option would be more gentle because it would tolerate a few malformed SA tags instead of terminating.

joshuak94: Hm that's annoying hahaha. Then let's leave this as it is!

joshuak94 (approved):

eldariont (changes requested):
All good - just one tiny wrong flag 😉

eldariont (approved):
Thanks! 👍

Issue #111, opened 2021-04-21
[FEATURE,FIX] Correct and improve VCF output


eldariont:
- [x] The VCF header line `CHROM\tPOS\tID\tREF\tALT\tQUAL\tFILTER\tINFO` is missing the preceding `#`.
- [x] Output INS calls based on the junctions (in `src/variant_detection/variant_output`)
- [x] Minimum variant size is currently hard-coded in `find_and_output_variants()`. We should use the command-line argument instead.
- [x] The coordinates returned by [record.reference_position](https://docs.seqan.de/seqan/3-master-dev/classseqan3_1_1sam__record.html#a959338eb203f0360499056f1530d7f4f) are 0-based but the POS field in VCF needs to be 1-based. Therefore, we need to add 1 to every coordinate in `find_and_output_variants()`.
- [x] Deletion junctions specify the last base `a` before and the first base `b` after the deletion. To output correct coordinates in VCF, we need to set `POS=a` and `END=b-1`, I believe.
- [x] Insertion junctions specify the last base `a` before and the first base `b=a+1` after the insertion. To output correct coordinates in VCF, we need to set `POS=a` and `END=a`, I believe.
- [x] The QUAL field of the output VCF is hard-coded to 60 for all calls. Instead, we can use the cluster size as a preliminary score for the calls.

PR #113, opened 2021-04-29
[FEATURE] Improve junction detection (insertion calling 1/2)

### LINKED ISSUE #94: [FEATURE] Improve junction detection
Firstly, there are problems with the current junction calling from SA tags:
- [x] Maximum overlap and gap between alignment segments is currently hard-coded in `analyze_aligned_segments()`. We should add new command-line arguments instead.
- [x] Method `get_reference_end()` of `AlignedSegment` class returns position after the last aligned base instead of last aligned base.
- [x] Overlapping alignment segments (`segment_a.get_query_end() > segment_b.get_query_start()`)  cause `seqan3::views::slice()` to crash
- [x] When reading SAM/BAM file, seqan returns 0-based coordinates. The SA tag, however, is stored as a string containing a 1-based position. We need to decrement this position `src/modules/sv_detection_methods/analyze_sa_tag_method.cpp`.

Secondly, insertions are currently not called from SA tags:
- [x] Produce insertion junctions in the split alignment case (in `src/modules/sv_detection_methods/analyze_sa_tag_method.cpp`)

eldariont:
This PR fixes #94 and adds the detection of junctions for insertions in the SA tag. The detected junctions are still not included in the VCF output which is the topic of #111.


eldariont:
Hi @SGSSGene, I rebased this PR after the recent merges and added another bugfix. For your review, it's probably the easiest to go through the commits one by one.

test/api/input_file_test.cpp
```diff
@@ -12,7 +12,8 @@ std::string const default_alignment_short_reads_file_path = DATADIR"paired_end_m
 std::string const default_alignment_long_reads_file_path = DATADIR"simulated.minimap2.hg19.coordsorted_cutoff.sam";
 std::filesystem::path const empty_output_path{};
 std::vector<detection_methods> const default_methods{cigar_string, split_read, read_pairs, read_depth};
-constexpr uint64_t sv_default_length = 30;
+constexpr uint64_t default_min_length = 30;
+constexpr uint64_t default_max_overlap = 10;
```
SGSSGene, line 17: I believe this should be `uint32_t`? Also why the swtich from `uin64_t` to `uint32_t` everywhere?

SGSSGene (changes requested):
I have seen you changed the types from uint64_t, to uint32_t.
I assume having reads longer than 4 billion is not a use case.
What about reads between 2billion and 4billion?
I have seen that the command line request `uint32_t`, but the algorithm uses internal many `int32_t` types. This could a source of errors (assuming 2billion long reads is a thing).

eldariont:
Hi @SGSSGene,
thanks for your comments!

> I assume having reads longer than 4 billion is not a use case.
> What about reads between 2billion and 4billion?

Yes, that's right. The longest sequencing reads I've ever heard of were Nanopore reads with lengths of a few million bases. Theoretically however, iGenVar could be applied to contigs from genome assemblies and these can get as big as entire chromosomes. The largest human chromosome is less than 300 million bases long but there are other species with larger chromosomes. However, I don't think there exist chromosomes with billions of bases. So all the numbers we are crunching in iGenVar should fit into `int32` (2,147,483,647) and `uint32` (4,294,967,295).

> I have seen that the command line request `uint32_t`, but the algorithm uses internal many `int32_t` types. This could a source of errors (assuming 2billion long reads is a thing).

That's a good point. I cannot think of a real-world scenario where our numbers would exceed 2 billion but it would be good to safe-guard against this case anyway. Our command-line parameters need to be unsigned (variant lengths cannot be negative) but we need to compute signed numbers in the code. How can this problem be solved?

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -57,29 +56,82 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
...
     {
         AlignedSegment current = aligned_segments[i-1];
         AlignedSegment next = aligned_segments[i];
         int32_t distance_on_read = next.get_query_start() - current.get_query_end();
-        // Check that there is neither an overlap nor a larger gap on the read
-        // TODO(eldarion): add command-line parameters for changing these currently hard-coded cutoffs
-        if (distance_on_read >= 0 && distance_on_read <= 10)
+        // Check that the overlap between two consecutive alignment segments
+        // of the read is lower than the given threshold
+        if (distance_on_read >= -static_cast<int32_t>(max_overlap))
```
eldariont, line 70: @SGSSGene: This line was the reason for my switch from `uint64_t` to `uint32_t`. I need to compare `distance_on_read` with `-max_overlap`. `distance_on_read` is an int32 while max_overlap was an uint64. To enable the comparison, I need to cast the unsigned integer to a signed integer. Casting `uint64_t` to `int32_t` seemed worse than casting `uint32_t` to `int32_t` even though this second conversion is still undefined for values larger than 2 billion. So I changed the parameter types from `uint64_t` to `uint32_t`.

Would it be a cleaner solution to cast the `uint32_t` to `int64_t` and compare this value to `distance_on_read`? If I interpret the standard correctly, then `distance_on_read` would be automatically converted to `int64_t`, too, for this comparison.

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -57,29 +56,82 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
...
+                }
+            }
+            int32_t distance_on_ref = mate2_pos - mate1_pos - 1;
+            // Check that the two consecutive alignment segments either
+            // map to different reference sequences (e.g. translocation, interspersed duplication),
+            // have a large distance on the reference (e.g. deletion, inversion, tandem duplication), or
+            // have a large distance on the read (e.g. insertion)
+            if (current.ref_name != next.ref_name ||
+                abs(distance_on_ref) >= min_length ||
+                distance_on_read >= min_length)
```
eldariont, line 115: @SGSSGene: You mentioned the potential problems caused by using unsigned and signed values together. In this line, we again have a comparison of signed with unsigned. Let's say `distance_on_read = -5` and `min_length = 30`. Would `distance_on_read` be converted to unsigned and take a wrong and very large value in this setting? If yes, what can be done about that?
Do we need to cast all unsigned parameters to a signed type every time we use them? If yes, I would rather use signed CLI parameters and add an additional check that their value is non-negative.

SGSSGene: Yes, this is an issue. If `distance_on_read` is negative it will be promoted to a `uint32_t`. Promotion rules are very weird.
So I think I would also do the same trick as above `distance_on_read >= static_cast<int32_t>(min_length)`
and a assert `assert(min_length <= std::numeric_limits<int32_t>::max())` at the beginning.

And adding a check in the CLI is also a good Idea. You could either use a `signed type`, or just add a check that the uint32v_t value is not larger than `std::numeric_limit<int32_t>::max()`. 


By the way, it might be better to use `std::abs` instead of `abs`. `std::abs` is the c++ version and will auto overload to `int`, `long`, `long long`, `float` etc.... In this particular case it doesn't matter, but `abs` will always call `abs(int)` and might force an implicit cast.

eldariont:
Hi @SGSSGene,

thanks for your explanations and comments. After some thinking on the matter, I came to the conclusion that it might be best to use signed instead of unsigned types for our CLI parameters. As [this](https://www.youtube.com/watch?v=wvtFGa6XJDU) 6-minute talk from CppCon nicely points out, mixing unsigned with signed types is always dangerous when you need to perform comparisons or mathematical operations on them. Yes, we can always cast unsigned to signed values and use asserts to guard against unexpected behavior. However, we always run the risk of missing comparisons that will behave unexpectedly without our knowing.
My intention of using unsigned types in the first place was to communicate that these parameters need to be non-negative. But in the talk above, Jon Kalb points out that this approach can also hide bugs in the code. Instead, he recommends to use documentation to communicate that a value should not be negative.

What is your take on this question? I just pushed a commit (78575ffbcd2dabeaacb5406110a4a69840cfe37a) and would like to know how you like this solution.

Cheers
David

SGSSGene:
I think using in this case a `signed integer` instead of `unsigned integer` is a good pick. But I would disagree on the reason why.
John Kolb most strongly argues that using `unsigned integer` can hide bugs. I disagree, I believe the same can happen when using `signed integer` it is just a different type of bug that gets visible. A second argument he makes is that API should communicate by documentation. I very strongly disagree with this one. The best API is an API that can not be used in a wrong way. Having condition on the parameters that are not expressed by code but by external documentation will increase the amount of bugs. It is hard to read and write manuals 100% correctly.

But!: I would still follow his advice on using `signed integer` instead of `unsigned integer`. In my opinion the flaws of `unsigned integer` and how ugly they make code with all the  necessary `static_cast`, makes the code so badly readable that it introduces more bugs, than it actually reduces by having a more rigorous data type.
I would also use a similar example as he used it `a - b`.
For `unsigned integer` it happens very fast that `a` might be smaller than `b` and we have an error.
If we use `signed integer` it is only a problem if `a` or `b` are around the maximum or minimum value which is very unlikely to occur.

SGSSGene (approved):
Looks good! Nice work :+1:

eldariont:
> A second argument he makes is that API constraints should be communicated by documentation. I very strongly disagree with this one. The best API is an API that can not be used in a wrong way. Having condition on the parameters that are not expressed by code but by external documentation will increase the amount of bugs. It is hard to read and write manuals 100% correctly.

Yes, I totally agree with you. I always try to design my code in such a way that it is understandable without any documentation. Unfortunately, using unsigned ints for this purpose leads to all these problems that eat up the advantage.

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -57,29 +56,82 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
...
-            junctions.push_back(std::move(new_junction));
+            int32_t mate1_pos;
+            if (current.orientation == strand::forward)
+            {
+                mate1_pos = current.get_reference_end();
+            }
+            else
+            {
+                mate1_pos = current.get_reference_start();
+            }
```
Irallia, line 80: You can shorten this, if you want:
```suggestion
            if (current.orientation == strand::forward)
                mate1_pos = current.get_reference_end();
            else
                mate1_pos = current.get_reference_start();
```

src/modules/sv_detection_methods/analyze_sa_tag_method.cpp
```diff
@@ -57,29 +56,82 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
...
+                // Correct position of mate 2 for overlapping alignment segments:
+                // Trim alignment of `next` segment at the start to remove overlap
+                if (distance_on_read < 0)
+                {
+                    mate2_pos = next.get_reference_start() - distance_on_read;
+                }
+                else
+                {
+                    mate2_pos = next.get_reference_start();
+                }
```
Irallia, line 93: and here

test/api/detection_test.cpp
```diff
@@ -212,61 +217,153 @@ TEST(junction_detection, retrieve_aligned_segments)
...
+}
+
+TEST(junction_detection, overlapping_segments)
+{
+    AlignedSegment aligned_segment1 {strand::forward, "chr1", 100, 60, std::vector<seqan3::cigar>{{20, 'M'_cigar_operation},
+                                                                                                  {30, 'S'_cigar_operation}}};
+    AlignedSegment aligned_segment2 {strand::forward, "chr1", 200, 60, std::vector<seqan3::cigar>{{15, 'S'_cigar_operation},
+                                                                                                  {35, 'M'_cigar_operation}}};
+    std::vector<AlignedSegment> aligned_segments{aligned_segment1,
+                                                 aligned_segment2};
```
Irallia, line 340:
```suggestion
    std::vector<AlignedSegment> aligned_segments{aligned_segment1, aligned_segment2};
```

test/api/detection_test.cpp
```diff
@@ -212,61 +217,153 @@ TEST(junction_detection, retrieve_aligned_segments)
...
         EXPECT_EQ(junctions_expected_res[i].get_read_name(), junctions_res[i].get_read_name()) << "Read names of junction " << i << " unequal";
         EXPECT_TRUE(junctions_expected_res[i] == junctions_res[i]) << "Junction " << i << " unequal\nMate 1 equal: "
-                                                                   << (junctions_expected_res[i].get_mate1() == junctions_res[i].get_mate1())
-                                                                   << "\nMate 2 equal: "
-                                                                   << (junctions_expected_res[i].get_mate2() == junctions_res[i].get_mate2())
-                                                                   << "\n";
+                                                                    << (junctions_expected_res[i].get_mate1() == junctions_res[i].get_mate1())
+                                                                    << "\nMate 2 equal: "
+                                                                    << (junctions_expected_res[i].get_mate2() == junctions_res[i].get_mate2())
+                                                                    << "\n";
```
Irallia, line 366: Now it's not aligned anymore..

test/data/mini_example/output_res.txt
```diff
@@ -5,10 +5,9 @@
 ##INFO=<ID=END,Number=1,Type=Integer,Description="End position of SV called.",Source="iGenVarCaller",Version="1.0">
 #CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO
 chr1	57	.	N	<DEL>	9	PASS	END=70;SVLEN=-13;SVTYPE=DEL
-chr1	99	.	N	<DEL>	1	PASS	END=125;SVLEN=-26;SVTYPE=DEL
+chr1	97	.	N	<DEL>	1	PASS	END=125;SVLEN=-28;SVTYPE=DEL
```
Irallia, line 8: This comes from 94+4 right?
`read027	2048	chr1	94	60	4M46S	*	0	0	ATTTGGATCTTGACTCTGGAAAACTTTTAACGCCGGGAATCGGTAGTCCT	*	AS:i:4	NM:i:0`

Irallia (changes requested):
Thanks for this improvement!
It's a little sad that the `test/data/mini_example/output_res.txt` still had bugs in it.
I hope I have understood the tests correctly so far. At least I have not found any mistakes in the content.
As always, there are only a few cosmetic corrections. 💅

test/data/mini_example/output_res.txt
eldariont: Yes, sort of. Here is the computation:
SAM and VCF are both 1-based so it's relatively easy. The first base of read027 is aligned to position 94 (that's what the POS field tells us). The alignment ends after 4 bases (CIGAR=4M46S) so that the last mapped base aligned to position 97. The first deleted base in the reference is therefore position 98. However, the VCF standard expects us to include a padding base left of the deletion. This padding base is position 97 which therefore is the POS field in the VCF.

Irallia (approved):
Perfect, thanks for the changes!
When you tidy up the commits, you're free to merge this :)
I think this thing is huge enough, that the -0.02% codecove is okay for now.

PR #120, opened 2021-05-12
[INFRA] Update coverage CI

eseiler:
A merge of https://github.com/seqan/app-template/pull/47

---
Old:
We also need to link the coverage config against the lib and the tests.

Also, I followed https://wiki.documentfoundation.org/Development/Lcov and try really hard not to run the tests before I got some baseline coverage, which is the same we do in seqan3.


Irallia (approved):
Thanks for the improvement ❤️

joshuak94 (approved):

joshuak94:
Looks good to me!

PR #121, opened 2021-05-12
[TEST] gcc11 support

### LINKED ISSUE #119: [TEST] Change tests, such that the tests also work for g++-11
> The enumeration_names is a map, so the order of keys is not fixed and may differ for each compiler/os/etc.

See https://github.com/seqan/iGenVar/pull/118#discussion_r629646440

eseiler:
Resolves #119 
Blocked by #120 (**i.e., only the `[TEST] gcc11 CI` commit is important**)

See also https://github.com/seqan/seqan3/pull/2642 and https://github.com/seqan/seqan3/pull/2644

* `seqan3::views::take_until_or_throw` is deprecated (but still in the detail namespace). It can be replaced by `std::views::take_while` with a negated predicate (`!is_digit` -> `is_digit`).
* seqan3 now fixes the order when printing enumeration_names. It is a map, so the order of elements when iterating is not guaranteed to be the same across compilers/OSes. We sort by values if they have an order, otherwise we sort by key. If two values are the same, we sort by key since keys are unique.


Irallia:
> Can't hurt to include [seqan/seqan3#2644](https://github.com/seqan/seqan3/pull/2644) if merged

As it is merged, you can update the seqan lib once more :)

Irallia (approved):
And now it is really nice ordered! :) ❤️

MitraDarja (approved):

Issue #127, opened 2021-05-19
[TEST] Fix VCF output for truvari


eldariont:
Benchmarking with truvari revealed two problems with our VCF output:
- [x] Genotype fields are missing: We need to add two columns (FORMAT and `<sample>`) where `<sample>` should be customizable by the user via a CLI parameter. Example:
```
#CHROM POS     ID        REF    ALT     QUAL FILTER INFO                              FORMAT      MYSAMPLE
20     14370   rs6054257 G      A       29   PASS   NS=3;DP=14;AF=0.5;DB;H2           GT          ./.
```

- [x] Connected to this, we need to define the format of the GT tag:
```
##FORMAT=<ID=GT,Number=1,Type=String,Description=\"Genotype\">
```

- [x] We need to add information about the reference contigs to the header. Example:
```
##contig=<ID=1,length=249250621>
##contig=<ID=2,length=243199373>
##contig=<ID=3,length=198022430>
```

Irallia:
>     * We need to add information about the reference contigs to the header. Example:
> 
> 
> ```
> ##contig=<ID=1,length=249250621>
> ##contig=<ID=2,length=243199373>
> ##contig=<ID=3,length=198022430>
> ```

Do we get this information from the `@SQ` tag in the SAM file? I found this in the VCF spec:
```text
It is recommended in VCF and required in BCF that the header includes tags describing the reference and contigsbacking the data contained in the file.  These tags are based on the SQ field from the SAM spec; all tags are optional(see the VCF example above)
```
(https://samtools.github.io/hts-specs/VCFv4.3.pdf)

eldariont:
> Do we get this information from the `@SQ` tag in the SAM file? I found this in the VCF spec:
> 
> ```
> It is recommended in VCF and required in BCF that the header includes tags describing the reference and contigsbacking the data contained in the file.  These tags are based on the SQ field from the SAM spec; all tags are optional(see the VCF example above)
> ```
> 
> (https://samtools.github.io/hts-specs/VCFv4.3.pdf)
Yes, we should be able to get this from the SQ field.

Irallia:
What about the `FORMAT` line?
We added this in our first benchmark test aswell: 
` "##FORMAT=<ID=GT,Number=1,Type=String,Description=\"Genotype\">"`
Do you think we can skip this?

eldariont:
> What about the `FORMAT` line?
> We added this in our first benchmark test aswell:
[...]

PR #128, opened 2021-05-19
[FEATURE, FIX] add contig to the vcf header 

Irallia:
Resolves part of #127 (contig to the vcf header)

We also update seqan3 here to get the fix for SAM/BAM header bugs: https://github.com/seqan/seqan3/issues/2600

### DIFF
--- a/include/variant_detection/variant_detection.hpp
+++ b/include/variant_detection/variant_detection.hpp
@@ -1,20 +1,32 @@
 #pragma once
 
 #include <seqan3/std/filesystem>    // for filesystem
+#include <map>
 #include <vector>
 
 #include "iGenVar.hpp"              // for struct cmd_arguments
 #include "structures/junction.hpp"  // for class Junction
 
+/*! \brief Reads the header of the input file. Checks if input file is sorted and reads the reference sequence
+ *         dictionary. Stores the reference sequence lengths in parameter `reference_lengths` and returns the list of
+ *         reference sequences.
+ *
+ * \param[in, out]  alignment_file - short or long reads input file
+ * \param[in, out]  references_lengths - reference sequence dictionary parsed from \@SQ header lines
+ */
+std::deque<std::string> read_header_information(auto & alignment_file,
+                                                std::map<std::string, int32_t> & references_lengths);
+
 /*! \brief Detects junctions between distant genomic positions by analyzing a short read alignment file (sam/bam). The
  *         detected junctions are stored in a vector.
  *
  * \param[in, out]  junctions - a vector of junctions
+ * \param[in, out]  references_lengths - reference sequence dictionary parsed from \@SQ header lines
  * \param[in]       args - command line arguments:\n
  *                         **args.alignment_short_reads_file_path** - short reads input file, path to the sam/bam file\n
  *                         **args.methods** - list of methods for detecting junctions
  *                            (0: cigar_string, 1: split_read, 2: read_pairs, 3: read_depth) - *default: all methods*\n
- * 
+ *
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
  *          We filter unmapped alignments, secondary alignments, duplicates and alignments with low mapping quality.
@@ -22,18 +34,22 @@
  *          For primary alignments, also the split read information is analyzed.
  */
 void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
+                                              std::map<std::string, int32_t> & references_lengths,
                                               cmd_arguments const & args);
 
 /*! \brief Detects junctions between distant genomic positions by analyzing a long read alignment file (sam/bam). The
  *         detected junctions are stored in a vector.
  *
  * \param[in, out]  junctions - a vector of junctions
+ * \param[in, out]  references_lengths - reference sequence dictionary parsed from \@SQ header lines
  * \param[in]       args - command line arguments:\n
  *                         **args.alignment_long_reads_file_path** - long reads input file, path to the sam/bam file\n
  *                         **args.methods** - list of methods for detecting junctions
  *                            (0: cigar_string, 1: split_read, 2: read_pairs, 3: read_depth) - *default: all methods*\n
- *                         **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp*\n
- *                         **args.max_overlap** - maximum overlap between alignment segments (expected to be non-negative) - *default: 10 bp*
+ *                         **args.min_var_length** - minimum length of variants to detect
+ *                            (expected to be non-negative) - *default: 30 bp*\n
+ *                         **args.max_overlap** - maximum overlap between alignment segments
+ *                            (expected to be non-negative) - *default: 10 bp*
  *
  *
  * \details Detects junctions from the CIGAR strings and supplementary alignment tags of read alignment records.
@@ -42,4 +58,5 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
  *          For primary alignments, also the split read information is analyzed.
  */
 void detect_junctions_in_long_reads_sam_file(std::vector<Junction> & junctions,
+                                             std::map<std::string, int32_t> & references_lengths,
                                              cmd_arguments const & args);
--- a/include/variant_detection/variant_output.hpp
+++ b/include/variant_detection/variant_output.hpp
@@ -10,34 +10,46 @@
 
 /*! \brief Detects genomic variants from junction clusters and prints them to output stream in VCF format.
  *
- * \param[in]       clusters    - input junction clusters
- * \param[in]       args        - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp* \n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in, out]  out_stream  - output stream
+ * \param[in]       references_lengths  - reference sequence dictionary parsed from \@SQ header lines
+ * \param[in]       clusters            - input junction clusters
+ * \param[in]       args                - command line arguments:\n
+ *                                        **args.min_var_length** - minimum length of variants to detect
+ *                                           (expected to be non-negative) - *default: 30 bp* \n
+ *                                        **args.max_var_length** - maximum length of variants to detect
+ *                                           (expected to be non-negative) - *default: 1,000,000 bp*\n
+ *                                        **args.max_tol_inserted_length** - longest tolerated inserted sequence at
+ *                                           non-INS SV types (expected to be non-negative) - *default: 5 bp*
+ * \param[in, out]  out_stream          - output stream
  *
  * \details Extracts genomic variants from given junction clusters.
- *          The quality of an SV is estimated based on the size of the cluster (i.e. the number of reads supporting the SV).
+ *          The quality of an SV is estimated based on the size of the cluster (i.e. the number of reads supporting the
+ *          SV).
  */
-void find_and_output_variants(std::vector<Cluster> const & clusters,
+void find_and_output_variants(std::map<std::string, int32_t> & references_lengths,
+                              std::vector<Cluster> const & clusters,
                               cmd_arguments const & args,
                               std::ostream & out_stream);
 
 
 /*! \brief Detects genomic variants from junction clusters and prints them in output file in VCF format.
  *
- * \param[in] clusters          - input junction clusters
- * \param[in] args              - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp*\n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in] output_file_path  - output file path
+ * \param[in]       references_lengths  - reference sequence dictionary parsed from \@SQ header lines
+ * \param[in]       clusters            - input junction clusters
+ * \param[in]       args                - command line arguments:\n
+ *                                        **args.min_var_length** - minimum length of variants to detect
+ *                                           (expected to be non-negative) - *default: 30 bp*\n
+ *                                        **args.max_var_length** - maximum length of variants to detect
+ *                                           (expected to be non-negative) - *default: 1,000,000 bp*\n
+ *                                        **args.max_tol_inserted_length** - longest tolerated inserted sequence at
+ *                                           non-INS SV types (expected to be non-negative) - *default: 5 bp*
+ * \param[in]       output_file_path    - output file path
  *
  * \details Extracts genomic variants from given junction clusters.
- *          The quality of an SV is estimated based on the size of the cluster (i.e. the number of reads supporting the SV).
+ *          The quality of an SV is estimated based on the size of the cluster (i.e. the number of reads supporting the
+ *          SV).
  */
 //!\overload
-void find_and_output_variants(std::vector<Cluster> const & clusters,
+void find_and_output_variants(std::map<std::string, int32_t> & references_lengths,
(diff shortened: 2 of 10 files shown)


include/variant_detection/variant_detection.hpp
```diff
@@ -10,6 +11,7 @@
  *         detected junctions are stored in a vector.
  *
  * \param[in, out]  junctions - a vector of junctions
+ * \param[in, out]  sn_ln_tag_tuple_set - a vector of @SQ header tags tuples: SN and LN
```
Irallia, line 14: I'm not sure about the name `sn_ln_tag_tuple_set` any better ideas?

include/variant_detection/variant_output.hpp
```diff
@@ -11,6 +11,7 @@
 /*! \brief Detects genomic variants from junction clusters and prints them to output stream in VCF format.
  *
  * \param[in]       clusters    - input junction clusters
+ * \param[in, out]  sn_ln_tag_tuple_set - a vector of @SQ header tags tuples: SN and LN
```
Irallia, line 14: I align this after naming desicion.

include/variant_detection/variant_output.hpp
```diff
@@ -21,13 +22,15 @@
  *          The quality of an SV is estimated based on the size of the cluster (i.e. the number of reads supporting the SV).
  */
 void find_and_output_variants(std::vector<Cluster> const & clusters,
+                              std::set<std::tuple<std::string, int32_t>> & sn_ln_tag_tuple_set,
                               cmd_arguments const & args,
                               std::ostream & out_stream);
```
SGSSGene, line 27: I would either put 'out' parameter at the beginning or at the end. Since you already have them at the beginning of the previous function, maybe move it to the first place. (doc also needs an update then)

```suggestion
void find_and_output_variants(std::set<std::tuple<std::string, int32_t>> & sn_ln_tag_tuple_set,
                              std::vector<Cluster> const & clusters,
                              cmd_arguments const & args,
                              std::ostream & out_stream);
```

include/variant_detection/variant_output.hpp
```diff
@@ -39,5 +42,6 @@ void find_and_output_variants(std::vector<Cluster> const & clusters,
  */
 //!\overload
 void find_and_output_variants(std::vector<Cluster> const & clusters,
+                              std::set<std::tuple<std::string, int32_t>> & sn_ln_tag_tuple_set,
                               cmd_arguments const & args,
```
SGSSGene, line 46: Same here, maybe have the output in the first position?

include/variant_parser/variant_record.hpp
```diff
@@ -66,16 +66,21 @@ class variant_header
...
     //!endcond
-    void print(stream_type & out_stream)
+    void print(std::set<std::tuple<std::string, int32_t>> & sn_ln_tag_tuple_set, stream_type & out_stream)
     {
         out_stream << "##fileformat=" << fileformat << '\n';
         out_stream << "##source=" << source << '\n';
+
+        for(std::tuple<std::string, int32_t> contig : sn_ln_tag_tuple_set)
+            out_stream << "##contig=<ID=" << std::get<0>(contig) << ",length=" << std::get<1>(contig) << ">\n";
+
```
SGSSGene, line 83: Structure bindings also work in this context. I think it makes the code more readable. It also avoids an unnecessary std::string copy, which depending on the length of the id doesn't really matter.
```suggestion
        for(auto const& [id, length] : sn_ln_tag_tuple_set)
            out_stream << "##contig=<ID=" << id << ",length=" << length << ">\n";

```

src/variant_detection/variant_detection.cpp
```diff
@@ -72,9 +83,18 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
             seqan3::debug_stream << num_good << " good alignments from short read file." << std::endl;
         }
     }
+
+    size_t i = 0;
+    for(std::string ref_id : ref_ids)
+    {
+        sn_ln_tag_tuple_set.insert({ref_id, std::get<0>(ref_id_info[i])});
```
SGSSGene, line 90: Avoid copying of std::string by using `std::string const&`.
Using emplace which will create the tuple in-place.
```suggestion
    for(std::string const& ref_id : ref_ids)
    {
        sn_ln_tag_tuple_set.emplace(ref_id, std::get<0>(ref_id_info[i]));
```

src/variant_detection/variant_detection.cpp
```diff
@@ -153,4 +170,12 @@ void detect_junctions_in_long_reads_sam_file(std::vector<Junction> & junctions,
             seqan3::debug_stream << num_good << " good alignments from long read file." << std::endl;
         }
     }
+
+    size_t i = 0;
+    for(std::string ref_id : ref_ids)
+    {
+        sn_ln_tag_tuple_set.insert({ref_id, std::get<0>(ref_id_info[i])});
```
SGSSGene, line 177: Same as above.
```suggestion
    for(std::string const& ref_id : ref_ids)
    {
        sn_ln_tag_tuple_set.emplace(ref_id, std::get<0>(ref_id_info[i]));
```

src/variant_detection/variant_detection.cpp
```diff
@@ -153,4 +170,12 @@ void detect_junctions_in_long_reads_sam_file(std::vector<Junction> & junctions,
...
+        ++i;
+    }
+    return;
```
SGSSGene, line 180:
```suggestion
```
No need :-)

src/variant_detection/variant_detection.cpp
```diff
@@ -72,9 +83,18 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
...
```
SGSSGene, line 93: No need.
```suggestion
```

SGSSGene (changes requested):

include/variant_detection/variant_detection.hpp
SGSSGene: Maybe just `sn_ln_tags`?
My thinking is that there is no need to encode the data type into the variable name (`set` and `tuple`).

That it is a `set` and `tuples` is not so relevant and also clear from the datatype.

If I see it correctly, SN means "sequence number" and LN is the length?
So maybe `seq_length` or `sequence_length`. (I don't like this one so much, also it seems like it is a `map` not a `set`).

SGSSGene:
Just want to make sure, you want to leave `sn_ln_tag_tuple_set` the way it is?

SGSSGene (changes requested):
Just want to make sure, you want to keep sn_ln_tag_tuple_set the way it is?

Irallia:
> Just want to make sure, you want to keep sn_ln_tag_tuple_set the way it is?

We decided to switch from a set to a map, since there can only be one length for an id. If not, we want to give a warning.
For the variable we made the decision for references_lenghts.

SGSSGene (approved):

src/variant_detection/variant_detection.cpp
```diff
@@ -70,9 +81,24 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
             seqan3::debug_stream << num_good << " good alignments from short read file." << std::endl;
         }
     }
+
+    size_t i = 0;
+    for(std::string const & ref_id : ref_ids)
+    {
+        int32_t ref_length = std::get<0>(ref_id_info[i]);
+        if (references_lengths.find(ref_id) != references_lengths.end())
```
eldariont, line 89: Here, you only check whether `ref_id` is already a key in `reference_lengths`. Shouldn't we also check whether `references_lengths[ref_id] == ref_length` and only output a warning if it is not?

include/variant_parser/variant_record.hpp
```diff
@@ -66,16 +66,21 @@ class variant_header
      *
      * \tparam stream_type - a stream to print the output to
      *
+     * \param[in, out] references_lengths - a vector of \@SQ header tags tuples: SN and LN
```
eldariont, line 69: see my comment above about the description of this parameter

eldariont: Also: Why is that [in, out] and not just [in]?

include/variant_detection/variant_output.hpp
```diff
@@ -10,34 +10,46 @@
 
 /*! \brief Detects genomic variants from junction clusters and prints them to output stream in VCF format.
  *
- * \param[in]       clusters    - input junction clusters
- * \param[in]       args        - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp* \n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in, out]  out_stream  - output stream
+ * \param[in, out]  references_lengths  - a vector of \@SQ header tags tuples: SN and LN
```
eldariont, line 13: Why is that parameter [in, out] and not [in]? The function does not change its value, does it?

include/variant_detection/variant_output.hpp
```diff
@@ -10,34 +10,46 @@
...
 
 /*! \brief Detects genomic variants from junction clusters and prints them in output file in VCF format.
  *
- * \param[in] clusters          - input junction clusters
- * \param[in] args              - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp*\n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in] output_file_path  - output file path
+ * \param[in, out]  references_lengths  - a vector of \@SQ header tags tuples: SN and LN
```
eldariont, line 36: Same question here.

eldariont (changes requested):
Thanks for this great work. Just a few small things :)

src/variant_detection/variant_detection.cpp
```diff
@@ -36,7 +47,7 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
         int32_t const ref_pos               = record.reference_position().value_or(-1); // 4: POS
         uint8_t const mapq                  = record.mapping_quality();                 // 5: MAPQ
         auto const header_ptr               = record.header_ptr();
-        auto const ref_ids = header_ptr->ref_ids();
+        ref_ids = header_ptr->ref_ids();
```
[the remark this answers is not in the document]
Irallia: I have tried a bit and now I think I have found a nicer solution.

include/variant_parser/variant_record.hpp
Irallia: Well seen that was a copy paste mistake.

include/variant_detection/variant_detection.hpp
```diff
@@ -1,39 +1,54 @@
...
 
 #include <seqan3/std/filesystem>    // for filesystem
+#include <map>
 #include <vector>
 
 #include "iGenVar.hpp"              // for struct cmd_arguments
 #include "structures/junction.hpp"  // for class Junction
 
+/*! \brief Reads the header of the input file. Checks if input file is sorted and reads the reference sequence
+ *         dictionary.
```
eldariont, line 11: I like your solution of obtaining the ref_ids in the `read_header_information` function. However, you should document all return values and their meanings.

```suggestion
/*! \brief Reads the header of the input file. Checks if input file is sorted and reads the reference sequence
 *         dictionary. Stores the reference sequence lengths in parameter `reference_lengths` and returns the list
 *         of reference sequences.
```

include/variant_detection/variant_output.hpp
```diff
@@ -10,34 +10,46 @@
 
 /*! \brief Detects genomic variants from junction clusters and prints them to output stream in VCF format.
  *
- * \param[in]       clusters    - input junction clusters
- * \param[in]       args        - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp* \n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in, out]  out_stream  - output stream
+ * \param[in]       references_lengths  - a vector of \@SQ header tags tuples: SN and LN
```
eldariont, line 13: You already changed the description of this parameter in `variant_detection.hpp`. Could you also do it here? :)

include/variant_detection/variant_output.hpp
```diff
@@ -10,34 +10,46 @@
...
 
 /*! \brief Detects genomic variants from junction clusters and prints them in output file in VCF format.
  *
- * \param[in] clusters          - input junction clusters
- * \param[in] args              - command line arguments:\n
- *                                **args.min_var_length** - minimum length of variants to detect (expected to be non-negative) - *default: 30 bp*\n
- *                                **args.max_var_length** - maximum length of variants to detect (expected to be non-negative) - *default: 1,000,000 bp*\n
- *                                **args.max_tol_inserted_length** - longest tolerated inserted sequence at non-INS SV types (expected to be non-negative) - *default: 5 bp*
- * \param[in] output_file_path  - output file path
+ * \param[in]       references_lengths  - a vector of \@SQ header tags tuples: SN and LN
```
eldariont, line 36: And here?

include/variant_parser/variant_record.hpp
```diff
@@ -66,16 +66,21 @@ class variant_header
      *
      * \tparam stream_type - a stream to print the output to
      *
-     * \param[in, out] out_stream - the output stream to print to
+     * \param[in]       references_lengths - sequence dictionary parsed from \@SQ header lines
```
eldariont, line 69:
```suggestion
     * \param[in]       references_lengths - reference sequence dictionary parsed from \@SQ header lines
```

test/api/input_file_test.cpp
```diff
@@ -190,7 +192,55 @@ TEST(input_file, long_read_sam_file_unsorted)
                        default_min_length,
                        default_max_overlap};
     EXPECT_THROW(detect_junctions_in_long_reads_sam_file(junctions_res,
+                                                         references_lengths,
                                                          args), seqan3::format_error);
 
     std::filesystem::remove(unsorted_sam_path);
 }
+
+TEST(input_file, short_and_long_read_sam_file_with_different_references_lenghts)
```
eldariont, line 201:
```suggestion
TEST(input_file, short_and_long_read_sam_file_with_different_references_lengths)
```

test/api/input_file_test.cpp
```diff
@@ -190,7 +192,55 @@ TEST(input_file, long_read_sam_file_unsorted)
...
+    std::map<std::string, int32_t> references_lengths{};
+
+    // Create a blank SAM file with SQ header tag with different length of one reference.
+    std::filesystem::path const tmp_dir = std::filesystem::temp_directory_path();     // get the temp directory
+    std::filesystem::path sam_path{tmp_dir/"unsorted.sam"};
+    std::ofstream sam{sam_path.c_str()};
+    sam << "@HD\tVN:1.6\tSO:coordinate\n" <<
+           "@SQ\tSN:testchr\tLN:1000\n" <<
+           "@SQ\tSN:testchr\tLN:1001\n" <<
+           "test1\t16\ttestchr\t1\t60\t10M\t=\t1\t0\tGCGCGCGCGC\tFFFFFFFFFF\n";
```
eldariont, line 226: Nice test case 👍 
However, it is more likely that we have two different SAM files that disagree concerning the length of testchr. It should be easy to create two SAMs instead of one and plug them into `detect_junctions_in_short_reads_sam_file` and `detect_junctions_in_long_reads_sam_file`, respectively.

eldariont (changes requested):
Thanks for the changes 👍 Just a few more small comments 🙈

include/variant_detection/variant_detection.hpp
Irallia: 🙈

test/api/input_file_test.cpp
Irallia: I have extended the testcase a bit to cover more edgecases.

eldariont (approved):
Great, thanks for your patience with me :) Looks good to merge after resolving the conflicts.

PR #132, opened 2021-05-28
[MISC] Update vcf header with empty genotype information

### LINKED ISSUE #127: [TEST] Fix VCF output for truvari

Irallia:
This is the follow up PR of #128 and thus resolves #127.


SGSSGene (approved):
Looks good to me :-)

include/variant_parser/variant_record.hpp
```diff
@@ -220,4 +224,6 @@ class variant_record
     float qual{};
     std::string filter{"PASS"};
     std::map<std::string, std::string> info{};
+    std::string format{"GT"};
+    std::string mysample{"./."};
```
eldariont, line 228: I would suggest to rename this field from `mysample` to `genotype` because that's what it contains. `mysample` is just a placeholder for the real sample name that the user can provide via the CLI.

eldariont (approved):
Nice 👍 Just one suggestion.

PR #134, opened 2021-05-31
[FEATURE] set number of decompression threads

### LINKED ISSUE #129: [FIX] Restrict number of decompression threads
By default, seqan uses all available cores to decompress a BAM file. This is not always desirable for iGenVar (e.g. on large servers where we do not want to consume dozens of cores for decompression).

Instead, we should add a CLI parameter for the number of decompression threads that defaults to 1.

Then, we can set the number of threads as described [here](https://docs.seqan.de/seqan/3-master-user/cookbook.html#setting_compression_threads).

See also https://github.com/seqan/seqan3/issues/1195#issuecomment-845975111

eldariont:
This PR fixes #129.

Changes:
- add CLI parameter --threads (default 1)
- use its value to set number of decompression threads
- rename -t (max_tol_inserted_length) to -q to associate --threads with -t

### DIFF
--- a/src/iGenVar.cpp
+++ b/src/iGenVar.cpp
@@ -2,6 +2,7 @@
 
 #include <map>
 
+#include <seqan3/contrib/stream/bgzf_stream_util.hpp>       // for bgzf_thread_count
 #include <seqan3/core/debug_stream.hpp>                     // for seqan3::debug_stream
 
 #include "modules/clustering/hierarchical_clustering_method.hpp"    // for the hierarchical clustering method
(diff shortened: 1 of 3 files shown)


src/iGenVar.cpp
```diff
@@ -1,6 +1,7 @@
 #include "iGenVar.hpp"
 
 #include <seqan3/core/debug_stream.hpp>                     // for seqan3::debug_stream
+#include <seqan3/contrib/stream/bgzf_stream_util.hpp>       // for bgzf_thread_count
```
MitraDarja, line 5: Alphabtical order?
```suggestion
#include <seqan3/contrib/stream/bgzf_stream_util.hpp> // for bgzf_thread_count
#include <seqan3/core/debug_stream.hpp> // for seqan3::debug_stream
```

MitraDarja (approved):
LGTM, just one minor thing. :)

Irallia (changes requested):
LGFM, just the one open comment from Mitra.

Irallia (approved):
Perfect! Thank you!

PR #135, opened 2021-06-01
[FEATURE] Add quality and cutoff parameters

Irallia:
For our very first benchmarks we realized, that these two parameters make a huge difference on our results:
- the filtering cutoff for the sv results (default 1 read)
- the cutoff for the hierarchical clustering (default distance of 10)

Thus I made two commits to add these values to our input parameters:
- [FEATURE] Add a quality input parameter for SVs (amount of supporting reads).
- [FEATURE] Add a clustering cutoff parameter for the hierarchical clustering

### DIFF
--- a/include/iGenVar.hpp
+++ b/include/iGenVar.hpp
@@ -19,6 +19,7 @@ struct cmd_arguments
     int32_t max_overlap = 10;
     int16_t threads = 1;
     int32_t min_qual = 1;
+    double hierarchical_clustering_cutoff = 10;
 };
 
 void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments & args);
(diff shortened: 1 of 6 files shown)


MitraDarja (approved):

include/iGenVar.hpp
```diff
@@ -16,6 +16,8 @@ struct cmd_arguments
     int32_t max_var_length = 1000000;
     int32_t max_tol_inserted_length = 5;
     int32_t max_overlap = 10;
+    int32_t sv_qual = 1;
```
eldariont, line 19: I would rename to `min_qual` to make clear that this is a cutoff.

src/iGenVar.cpp
```diff
@@ -72,6 +72,16 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
                       "Specify the maximum allowed overlap between two alignment segments. "
                       "This value needs to be non-negative.",
                       seqan3::option_spec::advanced);
+    parser.add_option(args.sv_qual, 'q', "sv_qual",
+                      "Specify the minimum quality (amout of supporting reads) of a strucutral variant. "
```
eldariont, line 76:
```suggestion
                      "Specify the minimum quality (amount of supporting reads) of a structural variant to be reported in the vcf output file."
```

The first two changes are typos that should be corrected also in `variant_output.hpp` and `iGenVar.hpp`. The last one is just a suggestion.

eldariont (changes requested):
Looks great 👍 Just two small comments.

eldariont (approved):
🎉

PR #137, opened 2021-06-02
[MISC] EnumValidator is not necessary anymore

Irallia:
After some changes in the argument parser of seqan3, the selfmade EnumValidator is not necessary anymore. 

~~As there could come some more changes in seqan3, we will wait for https://github.com/seqan/seqan3/pull/2695.~~


eseiler (approved):

joshuak94 (approved):
Looks good! I'm a bit confused though where the validation is happening now...?

Irallia:
> Looks good! I'm a bit confused though where the validation is happening now...?

SeqAn3 has its own validators, which can now handle enums (in our case just the default_validator is used). Its possible that the help_page message of the value_list_validator will get an update, than we could use that aswell and remove the text: "Value must be one of ...".

PR #140, opened 2021-06-07
[MISC] Restructure input parameters

### LINKED ISSUE #139: [MISC] Restructure input parameter
Suggestion:
```cpp
struct cmd_arguments
{
// Input / Output:
    /* -i */ std::filesystem::path alignment_short_reads_file_path{""};
    /* -j */ std::filesystem::path alignment_long_reads_file_path{""};
// VCF:
    /* -o */ std::filesystem::path output_file_path{};
    /* -s */ std::string vcf_sample_name{"GENOTYPE"};
    /* -a */ std::filesystem::path junctions_file_path{};
    /* -b */ std::filesystem::path clusters_file_path{};
// Others:
    /* -h - help - not part of the args struct */
    /* -v - verbose - not implementet yet */ 
    /* -t */ int16_t threads = 1;
// Methods:
    /* -d */ std::vector<detection_methods> methods{cigar_string, split_read, read_pairs, read_depth}; // default: all
    /* -c */ clustering_methods clustering_method{hierarchical_clustering};          // default: hierarchical clustering
    /* -r */ refinement_methods refinement_method{no_refinement};                    // default: no refinement
// SV specifications:
    /* -k */ int32_t min_var_length = 30;
    /* -l */ int32_t max_var_length = 1000000;
    /* -m */ int32_t max_tol_inserted_length = 5;
    /* -n */ int32_t max_overlap = 10;
    /* -q */ int32_t min_qual = 1;
// Clustering specifications:
    /* -w */ double hierarchical_clustering_cutoff = 10;
    /* x */
// Refinement specifications:
    /* y, z? */
};
```

Also change most one letter variables to the use of whole name parameters.

Irallia:
Rename single letters of input parameter and use full name instead of single letters for unusual parameters.

Resolves #139


test/cli/iGenVar_cli_test.cpp
```diff
@@ -428,7 +428,8 @@ TEST_F(iGenVar_cli_test, test_direct_methods_input)
 {
     cli_test_result result = execute_app("iGenVar",
                                          "-j", data(default_alignment_long_reads_file_path),
-                                         "--method 0 --method 1 --clustering_method 0 --refinement_method 0");
+                                         "--method cigar_string --method split_read "
+                                         "--clustering_method 0 --refinement_method 0");
```
eaasna, line 432: Not sure about these two strings. Shouldn't there be a comma in between?

eaasna (approved):
Looks good otherwise :)

test/cli/iGenVar_cli_test.cpp
Irallia: No, because its only one argument which is splitted into two lines.

test/api/detection_test.cpp
```diff
@@ -386,9 +386,10 @@ TEST(junction_detection, analyze_sa_tag)
...
                        std::filesystem::path{},
                        std::filesystem::path{},
-                       "GENOTYPE",
+                       "MYSAMPLE",
                        std::filesystem::path{},
                        std::filesystem::path{},
+                       1,
                        std::vector<detection_methods>{cigar_string, split_read, read_pairs, read_depth},
                        simple_clustering,
                        sVirl_refinement_method,
```
eldariont, line 395: I'm just now seeing this and it has nothing to do with this PR, but many of these `cmd_arguments` in the tests (detection_test and input_file_test) are missing some parameters at the end. Even though these apparently fall back to the default values we should probably set them explicitly.

eldariont (changes requested):
Thanks, this is a great improvement to out user-friendliness 👍 
Just a few small comments.

include/iGenVar.hpp
```diff
@@ -6,22 +6,33 @@
...
-    clustering_methods clustering_method{hierarchical_clustering};                              // default: hierarchical clustering method
-    refinement_methods refinement_method{no_refinement};                                        // default: no refinement
-    int32_t min_var_length = 30;
-    int32_t max_var_length = 1000000;
-    int32_t max_tol_inserted_length = 5;
-    int32_t max_overlap = 10;
-    int16_t threads = 1;
-    int32_t min_qual = 1;
-    double hierarchical_clustering_cutoff = 10;
+// Input:
```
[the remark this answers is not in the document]
Irallia: Jepp, but I think this is nicer, as you have a good overview, which letters are already used.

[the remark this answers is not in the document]
eldariont: Yes, almost nothing in the world comes without downsides 😉

test/api/detection_test.cpp
Irallia: Very good comment! I have completely adjusted the args.

eldariont (approved):
Thank you for the additional work with the args! This is looking great now 👍

PR #141, opened 2021-06-09
[Fix] Subsample large partitions

### LINKED ISSUE #130: [BUG] Subsample large partitions to speed up hierarchical clustering
The hierarchical clustering algorithm computes a pairwise distance matrix for the junctions in a given partition. The size of this matrix is quadratic and computation gets prohibitively slow for partitions with more than 1000 junctions.

This is a general problem of hierarchical clustering. Citing [Wikipedia](https://en.wikipedia.org/wiki/Hierarchical_clustering):

image

Nevertheless, we can use a workaround:
For large partitions, we can randomly choose a (smaller) subset of junctions and perform the actual clustering on the subsample.

Suggestion:
```
std::vector<Junction> subsample_partition(std::vector<Junction> const & partition, uint16_t const sample_size)
{
    std::vector<uint32_t> indices(partition.size());
    std::iota(indices.begin(), indices.end(), 0);
    std::random_device rd;
    std::mt19937 g(rd());
    std::shuffle(indices.begin(), indices.end(), g);
    indices.resize(sample_size);
    std::vector<Junction> subsample{};
    for (uint32_t index : indices)
    {
        subsample.push_back(partition[index]);
    }
    return subsample;
}
```

eldariont:
This PR introduces changes for speeding up the hierarchical clustering and therefore fixes #130.

Additionally, this PR fixes two other small problems:
- The setting of the number of decompression threads as introduced in #134 did not work. Probably, the reason is that the variable was set in the method `detect_variants_in_alignment_file()`. With this PR, the number of threads is set in `detect_junctions_in_short_reads_sam_file()` and `detect_junctions_in_long_reads_sam_file()` directly before opening the SAM/BAM files. This now worked in my tests.
- The `num_good` variable counts the number of valid alignments and is used for reporting the progress during runtime. This PR changes its type from uint16_t which was far too small for large SAM/BAM files to uint32_t.

### DIFF
--- a/src/modules/clustering/hierarchical_clustering_method.cpp
+++ b/src/modules/clustering/hierarchical_clustering_method.cpp
@@ -1,6 +1,9 @@
 #include "modules/clustering/hierarchical_clustering_method.hpp"
 
 #include <limits>                                                 // for infinity
+#include <random>                                                 // for random_device
+
+#include <seqan3/core/debug_stream.hpp>
 
 #include "fastcluster.h"                                          // for hclust_fast
 
@@ -104,11 +107,23 @@ int junction_distance(Junction const & lhs, Junction const & rhs)
     }
 }
 
+inline std::vector<Junction> subsample_partition(std::vector<Junction> const & partition, uint16_t const sample_size)
+{
+    assert(partition.size() >= sample_size);
+    std::vector<Junction> subsample{};
+    std::sample(partition.begin(), partition.end(), std::back_inserter(subsample),
+                sample_size, std::mt19937{std::random_device{}()});
+    return subsample;
+}
+
 std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const & junctions,
                                                     double clustering_cutoff)
 {
     auto partitions = partition_junctions(junctions);
     std::vector<Cluster> clusters{};
+    // Set the maximum partition size that is still feasible to cluster in reasonable time
+    // A trade-off between reducing runtime and keeping as many junctions as possible has to be made
+    const size_t max_partition_size = 200;
     for (std::vector<Junction> & partition : partitions)
     {
         size_t partition_size = partition.size();
@@ -116,6 +131,20 @@ std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const 
         {
             clusters.emplace_back(std::move(partition));
             continue;
+        }
+        if (partition_size > max_partition_size)
+        {
+            seqan3::debug_stream << "A partition exceeds the maximum size ("
+                                 << partition_size
+                                 << ">"
+                                 << max_partition_size
+                                 << ") and has to be subsampled. Representative partition member:\n["
+                                 << partition[0].get_mate1()
+                                 << "] -> ["
+                                 << partition[0].get_mate2()
+                                 << "]\n";
+            partition = subsample_partition(partition, max_partition_size);
+            partition_size = max_partition_size;
         }
         // Compute condensed distance matrix (upper triangle of the full distance matrix)
         std::vector<double> distmat ((partition_size * (partition_size - 1)) / 2);
--- a/src/iGenVar.cpp
+++ b/src/iGenVar.cpp
@@ -112,8 +112,6 @@ void detect_variants_in_alignment_file(cmd_arguments const & args)
     // Store junctions
     std::vector<Junction> junctions{};
     std::map<std::string, int32_t> references_lengths{};
-    // Set number of decompression threads
-    seqan3::contrib::bgzf_thread_count = args.threads;
 
     // short reads
     if (args.alignment_short_reads_file_path != "")
--- a/src/variant_detection/variant_detection.cpp
+++ b/src/variant_detection/variant_detection.cpp
@@ -53,6 +53,8 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
                                      seqan3::field::ref_offset, // 4: POS
                                      seqan3::field::mapq>;      // 5: MAPQ
 
+    // Set number of decompression threads
+    seqan3::contrib::bgzf_thread_count = args.threads;
     seqan3::sam_file_input alignment_short_reads_file{args.alignment_short_reads_file_path, my_fields{}};
 
     std::deque<std::string> const ref_ids = read_header_information(alignment_short_reads_file, references_lengths);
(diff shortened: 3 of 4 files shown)


src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -104,6 +106,22 @@ int junction_distance(Junction const & lhs, Junction const & rhs)
     }
 }
 
+std::vector<Junction> subsample_partition(std::vector<Junction> const & partition, uint16_t const sample_size)
+{
```
SGSSGene, line 110: I would throw in a `assert(partition.size() >= sample_size)` to indicate that a sample_size has to be larger or this algorithm doesn't work.

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -117,6 +135,11 @@ std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const
             clusters.emplace_back(std::move(partition));
             continue;
         }
+        if (partition_size > 200)
```
SGSSGene, line 138: Why is there a hard coded limit at 200?

SGSSGene (changes requested):
Just two minor things.

Also, I am checking right now on your issue with `seqan3::contrib::bgzf_thread_count = args.threads;`. From reading seqan code, this is expected. But I'm not sure if it is what we intended. But your fix is valid! The issue is that `bgzf_thread_count` is technically a different variable for every translation unit (.cpp file).

src/modules/clustering/hierarchical_clustering_method.cpp
eldariont: The limit sets the maximum partition size that is still feasible to cluster in a reasonable time. 200 has been a good trade-off between reducing runtime and keeping as many junctions as possible.

It would be nice to replace this hard-coded value with a nice CLI parameter but as it is not a very important value I decided against it. We already have more than 15 parameters and more will be added in the future.

Is there an alternative to adding CLI parameters for every hard-coded value?

SGSSGene: I see.
Maybe put a `const size_t max_partition_size = 200` with a comment infront of the `if` in line 139 and replace all the 200 with `max_partition_size`, so if someone wanted to change it in code, they could do it by only touching one line.
Also, do you think the user should be informed that the algorithm is behaving differently?
Maybe a `debug_stream << "The number of partition have exceeded " << max_partition_size << ". Applying subsampling to reduce number of partitions\n";`?

eldariont: Great suggestions 👍 I incorporated both of them together with a new test case to keep up the good coverage.

SGSSGene (approved):
This looks nice! :+1:

src/modules/clustering/hierarchical_clustering_method.cpp
joshuak94: I'm wondering if this assertion needs to be here... The only time the subsample is called is within the if statement below
```cpp
if (partition_size > max_partition_size)
    // ...
    partition = subsample_partition(partition, max_partition_size);
```
So it seems a bit redundant to check it first in the if statement and then again within the function itself?

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -104,11 +108,31 @@ int junction_distance(Junction const & lhs, Junction const & rhs)
...
+    std::random_device rd;
+    std::mt19937 g(rd());
+    std::shuffle(indices.begin(), indices.end(), g);
+    indices.resize(sample_size);
+    std::vector<Junction> subsample{};
+    for (uint32_t index : indices)
+    {
+        subsample.push_back(partition[index]);
+    }
+    return subsample;
```
joshuak94, line 125:
```suggestion
    std::vector<Junction> subsample{};
    std::sample(partition.begin(), partition.end(), std::back_inserter(subsample),
                sample_size, std::mt19937{std::random_device{}()});
    return subsample;
```

Is this not the same thing?

joshuak94 (changes requested):

src/modules/clustering/hierarchical_clustering_method.cpp
SGSSGene: True, but the `assert` will not be part of the program if it is build in Release.
I see it as a good way to document precondition that have to be hold met.

src/modules/clustering/hierarchical_clustering_method.cpp
SGSSGene: Intressting, I thought std::sample is c++20, but I am wrong, it is c++17!
@eldariont you should totally go with @joshuak94  solution. It is short and clean and just a very good solution. :+1:

joshuak94: Also wondering if it's worth it to make the function inline. @SGSSGene what do you think?

src/modules/clustering/hierarchical_clustering_method.cpp
eldariont: > True, but the `assert` will not be part of the program if it is build in Release.
> I see it as a good way to document precondition that have to be hold met.

I agree because I see each function as kind of a module that can be used and re-used multiple times. The `assert` is helpful to make sure that it is used in a correct manner.

src/modules/clustering/hierarchical_clustering_method.cpp
eldariont: Great suggestion, this is so much simpler and shorter 👍

joshuak94 (approved):
Looks good to me! I think you can get rid of the `#include <numeric>` and then merge!

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -1,6 +1,10 @@
 #include "modules/clustering/hierarchical_clustering_method.hpp"
 
 #include <limits>                                                 // for infinity
+#include <numeric>                                                // for iota
```
joshuak94, line 4:
```suggestion
```

joshuak94 (approved):

src/modules/clustering/hierarchical_clustering_method.cpp
SGSSGene: > Also wondering if it's worth it to make the function inline. @SGSSGene what do you think?

No, `inline` is not needed. The `inline` keyword isn't really related to the optimization technique inlining anymore (even though it might originated from there and is still mentioned in the standard).

Any function that could be inlined, might get inlined independently of the keyword `inline`. All functions that are members of classes automatically have the `inline` keyword set. The `inline` keyword actually allows you to break the "one definition rule"

In this case it might(very unlikely) actually do more harm than good.
Lets say you add another .cpp file and accidentally also define the function with the name and signature `subsample_partition(std::vector<Junction> const &, uint16_t)` inside. If no inline is given, you will get a multiple definition error while linking. If both are declared inline, the linker assumes they are the same function and will delete one of them.

Here you can see the same question+answer on Stackoverflow (careful, the wrong answer is the accepted one)
https://stackoverflow.com/a/157929/1293111

Issue #143, opened 2021-06-22
[FEATURE] Call (Tandem) Duplications


Irallia:
To become as good as SVim, we need to detect (Tandem) duplications. We have to extract them from the insertions by looking at their sequences.

1. Part: A characteristic of tandem duplications is the number of copies. We want to store this in the new variable int16_t tandem_dup_count. It is 0 if there is no duplication (e.g. insertion (novel element), deletion, inversion ...). 
2. Part: The VCF file includes some ALT lines, which are still missing.
3. Part: Call tandem duplications via CIGAR string
4. Part: Call tandem duplications via SA tag

PR #147, opened 2021-08-04
[FEATURE] Add tandem_dup_count to Junction

Irallia:
Resolves first part of #143.

A characteristic of tandem duplications is the number of copies. I store this in the new variable `int16_t tandem_dup_count`. It is 0 if there is no duplication (e.g. insertion (novel element), deletion, inversion ...).

### DIFF
--- a/include/structures/junction.hpp
+++ b/include/structures/junction.hpp
@@ -13,6 +13,7 @@ private:
     Breakend mate1{};
     Breakend mate2{};
     seqan3::dna5_vector inserted_sequence{};
+    size_t tandem_dup_count{};
     std::string read_name{};
 
 public:
@@ -29,8 +30,10 @@ public:
     Junction(Breakend the_mate1,
              Breakend the_mate2,
              auto const & the_inserted_sequence,
+             size_t the_tandem_dup_count,
              std::string the_read_name) : mate1{std::move(the_mate1)},
                                           mate2{std::move(the_mate2)},
+                                          tandem_dup_count{the_tandem_dup_count},
                                           read_name{std::move(the_read_name)}
     {
         if ((mate2.seq_name < mate1.seq_name) ||
@@ -60,6 +63,9 @@ public:
     */
     seqan3::dna5_vector get_inserted_sequence() const;
 
+    //! \brief Returns the number of tandem copies of this junction.
+    size_t get_tandem_dup_count() const;
+
     //! \brief Returns the name of the read giving rise to this junction.
     std::string get_read_name() const;
 };
@@ -70,6 +76,7 @@ inline constexpr stream_t operator<<(stream_t && stream, Junction const & junc)
     stream << junc.get_mate1() << '\t'
            << junc.get_mate2() << '\t'
            << junc.get_inserted_sequence().size() << '\t'
+           << junc.get_tandem_dup_count() << '\t'
            << junc.get_read_name();
     return stream;
 }
--- a/include/structures/junction.hpp
+++ b/include/structures/junction.hpp
@@ -81,15 +81,16 @@ inline constexpr stream_t operator<<(stream_t && stream, Junction const & junc)
     return stream;
 }
 
-/*! \brief A junction is smaller than another, if their first mate, second mate, or inserted sequence (in this order)
- *         is smaller than the corresponding element of the other junction.
+/*! \brief A junction is smaller than another, if their first mate, second mate, count of tandem duplications, or
+ *         inserted sequence (in this order) is smaller than the corresponding element of the other junction.
  *
  * \param lhs   left side junction
  * \param rhs   right side junction
  */
 bool operator<(Junction const & lhs, Junction const & rhs);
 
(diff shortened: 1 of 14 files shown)


src/structures/junction.cpp
```diff
@@ -27,11 +32,15 @@ bool operator<(Junction const & lhs, Junction const & rhs)
            : lhs.get_mate2() != rhs.get_mate2()
              ? lhs.get_mate2() < rhs.get_mate2()
              : lhs.get_inserted_sequence() < rhs.get_inserted_sequence();
+             // TODO (23.7.21, irallia): tandem_dup_count doesn't play a role here right?
```
Irallia, line 35: @joshuak94 @joergi-w If the amount of duplications is bigger or smaller between junctions, does not play a role for size comparsion right? Just for equality.

joergi-w: That would be weird... 
In general, the following must hold: If a < b and b < a are both false, then a = b. 
This would be violated if you omit the duplication count.

Irallia: Aaah, you mean a duplication is larger if it contains more duplications?
And thus, one junction is larger than another if this is the case.
Right, but we look before whether a junction is smaller from the position perspective (mate < mate). The question is, does it matter how high the tandem_dup_count is and if so where would you put that? Is that more relevant than the length of the inserted sequence?

joergi-w: I guess, I haven't yet fully understood what a junction is... 
Can two junctions with equal positions and sequence lengths possibly have a different duplication count? If yes, I would include it with the lowest priority.
My point is, if **any** of the fields in class `junction` differ, there must be a clear order (so you can only omit field comparisons that are indirectly determined, or that are omitted in the equality comparison as well).

include/structures/junction.hpp
```diff
@@ -29,8 +30,10 @@ class Junction
     Junction(Breakend the_mate1,
              Breakend the_mate2,
              auto const & the_inserted_sequence,
+             int16_t const & the_tandem_dup_count,
```
SGSSGene, line 33: No advantage of using `const&`
```suggestion
             int16_t the_tandem_dup_count,
```

include/structures/junction.hpp
```diff
@@ -29,8 +30,10 @@ class Junction
...
              std::string the_read_name) : mate1{std::move(the_mate1)},
                                           mate2{std::move(the_mate2)},
+                                          tandem_dup_count{std::move(the_tandem_dup_count)},
```
SGSSGene, line 36: integer are not movable. So it will copy anyways.
```suggestion
                                          tandem_dup_count{the_tandem_dup_count},
```

SGSSGene (changes requested):

src/structures/junction.cpp
Irallia: I have now thought about it again.
A junction describes a change in the read in comparison to the ref (e.g. insertion, duplication, deletion). It consists of two endpoints (breakends), a possibly inserted sequence and the name of the read and now also the number of duplications.
If we now compare two junctions, we want to know if they
1. share approximately the same positions (breakend 1 and 2), then
2. if they contain a similar number of duplications and
3. the inserted sequence is of similar length.

Important to note, for a novel element insertion the inserted sequence is the whole new sequence, for a duplication it is the duplicated sequence, this can be shorter if there are multiple duplications. Therefore, I now compare first on duplication and then on the sequence.
The content of the sequence is not compared yet, but that is still planned.

SGSSGene (approved):
:+1:

test/api/clustering_test.cpp
```diff
@@ -14,6 +14,7 @@ int32_t const chrom1_position2 = 94734377;
 int32_t const chrom1_position3 = 112323345;
 std::string const chrom2 = "chr2";
 int32_t const chrom2_position1 = 234432;
+int16_t tandem_dup_count = 0;
```
joergi-w, line 17: use `const`?
However, it would be good to write at least one test with more than 0 tandem duplications...

joergi-w (changes requested):
Looks good so far! Two remarks:

src/structures/cluster.cpp
```diff
@@ -70,6 +70,30 @@ Breakend Cluster::get_average_mate2() const
...
+        } else {
+            sum_counts += members[i].get_tandem_dup_count();
+        }
+    }
+    // If two thirds of the junctions have a 0 tandem_dup_count, than its probably no tandem duplication.
+    if (amount_zero_counts > std::round(static_cast<double>(members.size()) / 3) )
+    {
+        return 0;
+    } else {
+        int32_t average_count = std::round(static_cast<double>(sum_counts) / members.size());
```
joergi-w, line 92: The type should match the return type of the function.
```suggestion
        uint16_t average_count = (0.5 + sum_counts) / members.size();
```

or
```suggestion
        return (0.5 + sum_counts) / members.size();
```

joergi-w (changes requested):
some comments on data types again...

joergi-w (approved):
Thanks!

Issue #20, opened 2020-10-05
[FEATURE] Add verbose option to detect_beackends & find_deletions


Irallia:
We should discuss, if we want a verbose option. There are some commented out prints that you might want to use here.

PR #157, opened 2021-09-16
[FEATURE] Add verbose option

Irallia:
Resolves partly #20 
We add a verbose option here, without different values, just as a flag.

### DIFF
--- a/src/iGenVar.cpp
+++ b/src/iGenVar.cpp
@@ -49,6 +49,9 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments 
     parser.add_option(args.threads, 't', "threads",
                       "Specify the number of decompression threads used for reading BAM files.",
                       seqan3::option_spec::standard);
+    parser.add_flag(gVerbose, 'v', "verbose",
+                    "If you set this flag, we provide additional details about what iGenVar does. The detailed output "
+                    "is printed in the standard error.");
 
     // Options - Optional output:
     parser.add_option(args.junctions_file_path, 'a', "junctions",
--- a/test/cli/iGenVar_cli_test.cpp
+++ b/test/cli/iGenVar_cli_test.cpp
@@ -48,6 +48,9 @@ std::string const help_page_part_1
     "    -t, --threads (signed 16 bit integer)\n"
     "          Specify the number of decompression threads used for reading BAM\n"
     "          files. Default: 1.\n"
+    "    -v, --verbose\n"
+    "          If you set this flag, we provide additional details about what\n"
+    "          iGenVar does. The detailed output is printed in the standard error.\n"
 };
 
 std::string const help_page_part_2
@@ -162,10 +165,6 @@ std::string expected_res_empty
 std::string expected_err_default_no_err
 {
     "Detect junctions in long reads...\n"
-    "INS: chr21\t41972615\tForward\tchr21\t41972616\tForward\t1681\t0\tm2257/8161/CCS\n"
-    "BND: chr21\t41972615\tReverse\tchr22\t17458415\tReverse\t2\t0\tm41327/11677/CCS\n"
-    "BND: chr21\t41972616\tReverse\tchr22\t17458416\tReverse\t0\t0\tm21263/13017/CCS\n"
-    "BND: chr21\t41972616\tReverse\tchr22\t17458416\tReverse\t0\t0\tm38637/7161/CCS\n"
     "Start clustering...\n"
     "Done with clustering. Found 2 junction clusters.\n"
     "No refinement was selected.\n"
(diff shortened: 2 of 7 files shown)


eaasna (approved):
LGTM

include/modules/clustering/hierarchical_clustering_method.hpp
```diff
@@ -37,10 +37,13 @@ int junction_distance(Junction const & lhs, Junction const & rhs);
...
  * \param[in] clustering_cutoff - distance cutoff for clustering
+ * \param[in] verbose           - verbose option
  *
  * \details For the algorithms we use the library hclust.
  * \see https://lionel.kr.hs-niederrhein.de/~dalitz/data/hclust/ (last access 01.06.2021).
  */
-std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const & junctions, double clustering_cutoff);
+std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const & junctions,
+                                                    double clustering_cutoff,
+                                                    bool const verbose);
```
joergi-w, line 49: I would remove `const` here. Since `bool` (also the `double` above) is passed by copying the value, you will not modify the caller's object anyways.
```suggestion
                                                    bool verbose);
```

include/modules/sv_detection_methods/analyze_cigar_method.hpp
```diff
@@ -36,4 +37,5 @@ void analyze_cigar(std::string const & read_name,
                    std::vector<seqan3::cigar> & cigar_string,
                    seqan3::dna5_vector const & query_sequence,
                    std::vector<Junction> & junctions,
-                   int32_t const min_length);
+                   int32_t const min_length,
+                   bool const verbose);
```
joergi-w, line 41: same here (and in many other places of this PR)

src/iGenVar.cpp
```diff
@@ -49,6 +49,9 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
     parser.add_option(args.threads, 't', "threads",
                       "Specify the number of decompression threads used for reading BAM files.",
                       seqan3::option_spec::standard);
+    parser.add_flag(args.verbose, 'v', "verbose",
+                    "If you set this flag to true, we provide additional details about what iGenVar does. The detailed "
+                    "output is printed in the standard output.");
```
joergi-w, line 54: Not true, you use debug_stream, which prints to stderr.
And stderr is good, because otherwise the output would be intertwined with the vcf output (which goes to stdout).

test/cli/iGenVar_cli_test.cpp
```diff
@@ -48,6 +48,10 @@ std::string const help_page_part_1
     "    -t, --threads (signed 16 bit integer)\n"
     "          Specify the number of decompression threads used for reading BAM\n"
     "          files. Default: 1.\n"
+    "    -v, --verbose\n"
+    "          If you set this flag to true, we provide additional details about\n"
+    "          what iGenVar does. The detailed output is printed in the standard\n"
```
joergi-w, line 53: please adapt

joergi-w (changes requested):
I'm not sure, if I should agree or disagree with this... So I share my thoughts:

As you see, the verbose variable is needed in every corner of the program and you have to change basically each and every function signature of the software. In my opinion, this is ***THE*** situation to use a public global variable. Your PR would shrink to a few lines, the future maintenance would be simpler, and you would avoid passing the boolean by copy(!) to each function. On the other hand I know that it is controversially discussed whether global variables should be used at all... Find some arguments [here](https://www.tutorialspoint.com/Why-should-we-avoid-using-global-variables-in-C-Cplusplus) and [here](https://www.learncpp.com/cpp-tutorial/why-non-const-global-variables-are-evil/).

src/iGenVar.cpp
SGSSGene: Instead of `If you set this flag to true,` just `If you set this flag`, because the call is `./program --verbose` without `true` as a parameter.

include/modules/clustering/hierarchical_clustering_method.hpp
SGSSGene: I slightly disagree. Putting `const` everywhere you can, never hurts (assuming you don't want to change the value of the variable ;-) ).
`const` is just a helper to avoid programming mistakes.
(On the other hand, I am also to lazy to put `const` everywhere)

joergi-w: Yeah, that's true, it never hurts. 
I'm just a friend of consistency and if the style in the line above is different, I get alerted :)

SGSSGene: I see, you truly life by the quote:  "Always code as if the guy who ends up maintaining your code will be a violent psychopath who knows where you live. Code for readability."! :1st_place_medal:

joergi-w (changes requested):
Looks good, thanks! Please check the two unresolved comments about the argument parser description of the `-v` flag.

Irallia:
I should not work too late, then I always miss something :sweat_smile:
Have now adjusted everything, I hope now nothing is missing anymore.

joergi-w (approved):

PR #158, opened 2021-09-16
[FEATURE] improve performance by eldariont

Irallia:
This PR replaces https://github.com/seqan/iGenVar/pull/142 

> This PR contains several changes to
> 
>     * the variant detection from split alignments,
> 
>     * the hierarchical clustering and
> 
>     * the variant output logic
> 
> 
> with the goal of improving the performance of our caller in the HG002 SV benchmark.
> 
> Currently, this is still a draft and we should measure the change in benchmark results it causes (after merging #141). I also need to fix the tests.

# Performance improvement:
before
[CallerComparisonPlot.pdf](https://github.com/seqan/iGenVar/files/7184928/CallerComparisonPlot.pdf)

after
[CallerComparisonPlot.pdf](https://github.com/seqan/iGenVar/files/7184943/CallerComparisonPlot.pdf)

# Parameter changes:
hierarchical_clustering_cutoff:
hierarchical_clustering_cutoff results all
max_tol_deleted_length:
max_tol_deleted_length results all
partition_max_distance:
partition_max_distance results all


src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -87,25 +91,54 @@ std::vector<std::vector<Junction>> split_partition_based_on_mate2(std::vector<Ju
...
+                                        lhs.get_mate2().position;
+            int32_t rhs_directed_size = rhs.get_inserted_sequence().size() +
+                                        rhs.get_mate1().position -
+                                        rhs.get_mate2().position;
+            // lhs and rhs have the same type (either deletion/inversion or insertion)
+            if ((lhs_directed_size < 0 && rhs_directed_size < 0) ||
+                (lhs_directed_size > 0 && rhs_directed_size > 0))
+            {
+                double position_distance = std::abs(lhs.get_mate1().position - rhs.get_mate1().position) / 1000.0;
+                // TODO (irallia 01.09.2021): std::abs((int)(lhs.get_tandem_dup_count() - rhs.get_tandem_dup_count()))
```
SGSSGene, line 118: What about this todo?

src/variant_detection/variant_output.cpp
```diff
@@ -32,10 +32,11 @@ void find_and_output_variants(std::map<std::string, int32_t> & references_length
...
                     if (mate1.orientation == strand::forward)
                     {
-                        int32_t distance = mate2_pos - mate1_pos;
+                        int32_t distance = mate2_pos - mate1_pos - 1;
+                        int32_t sv_length = insert_size - distance;
                         //Deletion
-                        if (distance >= args.min_var_length &&
-                            distance <= args.max_var_length &&
+                        if (sv_length <= -args.min_var_length &&
+                            sv_length >= -args.max_var_length &&
```
SGSSGene, line 39: These lines are very confusing... is sv_length a negative number?
I would suggest to write:
```suggestion
                        if (-sv_length >= args.min_var_length &&
                            -sv_length <= args.max_var_length &&
```

src/variant_detection/variant_output.cpp
```diff
@@ -45,15 +46,16 @@ void find_and_output_variants(std::map<std::string, int32_t> & references_length
                             tmp.add_info("SVTYPE", "DEL");
                             // Increment position by 1 because VCF is 1-based
                             tmp.set_pos(mate1_pos + 1);
-                            tmp.add_info("SVLEN", std::to_string(-distance + 1));
+                            tmp.add_info("SVLEN", std::to_string(sv_length));
                             // Increment end by 1 because VCF is 1-based
                             // Decrement end by 1 because deletion ends one base before mate2 begins
```
SGSSGene, line 51: These two comments seem to be outdated. There is no increment/decrement

SGSSGene (changes requested):
Looks good, just some small stuff.

src/variant_detection/variant_output.cpp
Irallia: Yes, but only because they cancel each other out.

SGSSGene: :facepalm: yes, they cancel each other out, perfectly good comment :-)

src/modules/clustering/hierarchical_clustering_method.cpp
Irallia: Thats a todo for my next tandem duplication PR :)

SGSSGene (approved):
nice :-)

Irallia:
@joergi-w : I would ignore the codecoverage loss. We can also go through parts of this PR on Wednesday if some things are not clear to you. It's a lot of stuff...
It is the code that was left from @eldariont . And since it actually brings some improvement, I rebased and added it.

include/modules/clustering/hierarchical_clustering_method.hpp
```diff
@@ -32,15 +37,18 @@ std::vector<std::vector<Junction>> split_partition_based_on_mate2(std::vector<Ju
  * \param[in] lhs - left side junction
  * \param[in] rhs - right side junction
  */
-int junction_distance(Junction const & lhs, Junction const & rhs);
+double junction_distance(Junction const & lhs, Junction const & rhs);
```
joergi-w, line 40: missing `\returns` in doxygen documentation (+ 3x in this file)

include/iGenVar.hpp
```diff
@@ -27,11 +27,13 @@ struct cmd_arguments
...
     /* -k */ int32_t min_var_length = 30;
     /* -l */ int32_t max_var_length = 100000;
-    /* -m */ int32_t max_tol_inserted_length = 5;
+    /* -m */ int32_t max_tol_inserted_length = 50;
+    /* -e */ int32_t max_tol_deleted_length = 50;
     /* -n */ int32_t max_overlap = 10;
     /* -q */ int32_t min_qual = 1;
 // Clustering specifications:
-    /* -w */ double hierarchical_clustering_cutoff = 100;
+    /* -p */ int32_t partition_max_distance = 1000;
```
joergi-w, line 35: If the user must pass non-negative values, just use an unsigned type. This automatically puts `(unsigned 32 bit integer)` to the help screen and prints a helpful error message if a user passes a negative value nonetheless.
```suggestion
    /* -p */ uint32_t partition_max_distance = 1000;
```

src/iGenVar.cpp
```diff
@@ -94,6 +94,10 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
                       "Specify what should be the longest tolerated inserted sequence at sites of non-INS SVs. "
                       "This value needs to be non-negative.",
                       seqan3::option_spec::advanced);
+    parser.add_option(args.max_tol_deleted_length, 'e', "max_tol_deleted_length",
+                      "Specify what should be the longest tolerated deleted sequence at sites of non-DEL SVs. "
+                      "This value needs to be non-negative.",
```
joergi-w, line 99: this extra comment wouldn't be necessary with unsigned parameter variable

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -25,13 +26,14 @@ std::vector<std::vector<Junction>> partition_junctions(std::vector<Junction> con
         {
             if (junction.get_mate1().seq_name != current_partition.back().get_mate1().seq_name ||
                 junction.get_mate1().orientation != current_partition.back().get_mate1().orientation ||
-                abs(junction.get_mate1().position - current_partition.back().get_mate1().position) > 50)
+                abs(junction.get_mate1().position - current_partition.back().get_mate1().position)
```
joergi-w, line 29: doesn't this need `std::abs`? (I'm not sure, but if it works on every machine - fine!)
Update: at least below you use `std::`, so decide on one version please

src/variant_detection/variant_detection.cpp
```diff
@@ -90,11 +90,13 @@ void detect_junctions_in_short_reads_sam_file(std::vector<Junction> & junctions,
                     break;
             }
         }
-
-        num_good++;
-        if (num_good % 1000 == 0)
+        if (gVerbose)
         {
-            seqan3::debug_stream << num_good << " good alignments from short read file." << std::endl;
+            num_good++;
```
joergi-w, line 95: avoid a copy:
```suggestion
            ++num_good;
```

src/variant_detection/variant_detection.cpp
```diff
@@ -174,10 +176,13 @@ void detect_junctions_in_long_reads_sam_file(std::vector<Junction> & junctions,
             }
         }
 
-        num_good++;
-        if (num_good % 1000 == 0)
+        if (gVerbose)
         {
-            seqan3::debug_stream << num_good << " good alignments from long read file." << std::endl;
+            num_good++;
```
joergi-w, line 181: same here

joergi-w (changes requested):

include/iGenVar.hpp
Irallia: I opened an issue for that: https://github.com/seqan/iGenVar/issues/161

src/iGenVar.cpp
Irallia: -> https://github.com/seqan/iGenVar/issues/161

joergi-w (approved):

Issue #151, opened 2021-09-01
Read a genome


joergi-w:
Provide a new argument for the ArgumentParser (`-r`?) to read a filename. Read the file and store the contained genome.

PR #162, opened 2021-09-23
Compute raw activity profile

### LINKED ISSUE #152: Calculate a raw activity profile
Use a reference sequence and aligned reads:
Count mismatches, indels and soft clips for each reference position. 
Transform these into a measure of entropy → activity score.

Result: Wave function of activity per genome position.

joergi-w:
Resolves #152: Scans a sam file and creates an activity vector per reference genome.

Solves also part of #151, as it introduces a parameter for a genome filename (the file is not yet read). If a genome file name is given, iGenVar switches to the SNP/Indel calling mode.

There are no tests yet for this functionality.


Irallia:
This looks like you need a test for the new error message. I just reordered all the CLI tests: https://github.com/seqan/iGenVar/pull/164 maybe wait with adding the test for this PR?
But you can already ask for the first review I think :)

joergi-w:
Thanks! I'll wait for the pending PR and then update this one on Wednesday.

src/iGenVar.cpp
```diff
@@ -35,6 +36,11 @@ void initialize_argument_parser(seqan3::argument_parser & parser, cmd_arguments
                       "Input long read alignments in SAM or BAM format (PacBio, Oxford Nanopore, ...).",
                       seqan3::option_spec::standard,
                       seqan3::input_file_validator{{"sam", "bam"}} );
+    parser.add_option(args.genome_file_path,
+                      'g', "input_genome",
+                      "Input the reference genome in FASTA or FASTQ format.",
+                      seqan3::option_spec::standard,
+                      seqan3::input_file_validator{{"fasta", "fa", "fastq"}} );
```
SGSSGene, line 43: You could add all valid file extensions (we really need some kind of defaults in seqan3)
```suggestion
                      seqan3::input_file_validator{{"fasta", "fa", "fna", "ffn", "ffa", "frn", "fastq", "fq"}} );
```

SGSSGene (approved):
Could add a few more file extensions, but looks good to go.

src/iGenVar.cpp
Irallia: I have seen `fna` before, the others should be double checked, but I think this is a good idea.
I found for FASTA in the german wiki: `.fa`, `.mpfa`, `.fna`, `.fsa`, `.fasta` and in the english wiki: .`fasta`, `.fna`, `.ffn`, `.faa`, `.frn`, `.fa`.
I think we don't have to support all of them, we can expect the user to customize his file extension, but a few like `fna` and `fq` we could possibly add.

src/variant_detection/snp_indel_detection.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                      seqan3::field::cigar>;     // 6: CIGAR
+
+    seqan3::sam_file_input reads_file{reads_filename, sam_fields{}};
+    auto const & header = reads_file.header();
+
+    // Check that the file is sorted.
+    if (header.sorting != "coordinate")
+        throw seqan3::format_error{"ERROR: Input file must be sorted by coordinate (e.g. samtools sort)"};
+
+    // Prepare one activity vector per reference sequence.
```
Irallia, line 27: Question: Are we talking about different references that were aligned to (with different sam files), or just individual fasta files for e.g. the individual chromosomes?

src/variant_detection/snp_indel_detection.cpp
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        // Track the current position within the read.
+        unsigned read_pos = 0;
+
+        // Step through the CIGAR string.
+        for (auto && [length, operation] : record.cigar_sequence())
+        {
+            // Skip long variants.
+            if (length >= 30) // TODO: use min_length parameter
```
Irallia, line 57: Since there are several of us working on this project, I have become accustomed to the following notation for TODOs:
```suggestion
            if (length >= 30) // TODO (joergi-w 30.09.2021): use min_length parameter
```
Then you know who to ask when you stumble across a TODO and also how old it is.

src/variant_detection/snp_indel_detection.cpp
Irallia: Aaah I think its the second one, as you use ref_id below. :)

src/variant_detection/snp_indel_detection.cpp
```diff
@@ -0,0 +1,94 @@
...
+                    break;
+                }
+                case 'D': // deletion
+                {
+                    for (unsigned idx = 0; idx < length; ++idx)
+                        ++activity[ref_id][ref_pos + idx];
+                    ref_pos += length;
+                    break;
+                }
+                default: // match or mismatch
```
Irallia, line 82: This default also coveres `N` (skipped region),`H` (hard clipping), `P` (padding). So I think this case is fine, you should maybe just adjust the comment a bit.

test/cli/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ cmake_minimum_required (VERSION 3.11)
 add_cli_test(iGenVar_cli_test.cpp)
 add_dependencies (iGenVar_cli_test "iGenVar")
 target_use_datasources (iGenVar_cli_test FILES simulated.minimap2.hg19.coordsorted_cutoff.sam)
+target_use_datasources (iGenVar_cli_test FILES mini_example_reference.fasta)
```
Irallia, line 6: Oh, I think this file needs to be updated, because I think I only updated the explanation fasta the last time I expanded my example.... 🙈

test/cli/iGenVar_cli_test.cpp
```diff
@@ -293,6 +295,33 @@ TEST_F(iGenVar_cli_test, test_intermediate_result_output)
...
+                                     "0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,13,2,2,2,2,2,2,2,2,"
+                                     "2,2,2,2,2,2,2,67,3,3,3,67,1,1,1,1,1,1,1,1,1,1,1,1,43,0,0,0,0,0,0,0,0,0,0,0,0,0,"
+                                     "0,2,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,4,4,4,4,4,4,4,4,4,4,4,4,4,4,0,0,0,"
+                                     "0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,55,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,1,"
+                                     "1,1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,"
+                                     "0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,52,"
+                                     "0,0,0,0,0,0,0,0,0,0]\n"
+                                     "Start clustering...\n"
+                                     "Done with clustering. Found 0 junction clusters.\n"
+                                     "No refinement was selected.\n";
```
Irallia, line 319: Question for the next meeting: Do we have our own clustering method here or do we use the existing one? Do we need the refinement for SNPs at all?

Irallia (changes requested):
Many, many thanks for this PR!
Do not be put off by the many comments. They are mainly small things and questions.

src/iGenVar.cpp
joergi-w: Good point, thanks! 
I agree that it should be a constant in SeqAn for sequence files in general.

include/variant_detection/snp_indel_detection.hpp
```diff
@@ -0,0 +1,9 @@
+#pragma once
+
+#include <seqan3/std/filesystem>
+
+/*!
+ * \brief Detect Single Nucleotide Polymorphisms (SNPs) and short deletions and insertions.
+ * \param reads_filename The file path where to find the sequenced reads.
```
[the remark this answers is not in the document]
joergi-w: Why do we put this hyphen? I don't find the meaning in the Doxygen manual for [\param](https://www.doxygen.nl/manual/commands.html#cmdparam). I see it just ends up as the beginning of the description string.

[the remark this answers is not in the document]
joergi-w: html_doc
pdf_doc

src/variant_detection/snp_indel_detection.cpp
joergi-w: In one sam file there can be linked various reference sequences. Whether they come from a single or multiple sequence files, we need to discuss at some point, but here I just use the ref ids from the sam file.

include/variant_detection/snp_indel_detection.hpp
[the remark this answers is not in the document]
Irallia: We once decided on this style, either we decide to change in general, or stay consistent. -> https://github.com/seqan/iGenVar/wiki/Coding-Guide
I find it clearer in the first image, but yes it looks a bit unnecessary in the second image. So if you think the change is important, then we should discuss it again and change it in a separate PR.

test/cli/CMakeLists.txt
Irallia: I will update it in my next PR, where I worked on this anyways

Irallia (approved):
Thanks! ❤️

PR #164, opened 2021-09-27
[MISC] Remove some warnings and update parser.

### LINKED ISSUE #161: [MISC] Update iGenVar parser - pass non-negative values and leave the handling to the parser
As in https://github.com/seqan/iGenVar/pull/158#discussion_r713710284, we should use unsigned types for passed non-negative values and leave the error handling to the argument parser.
> If the user must pass non-negative values, just use an unsigned type. This automatically puts (unsigned 32 bit integer) to the help screen and prints a helpful error message if a user passes a negative value nonetheless.

Irallia:
Resolves #161 

When you build iGenVar with `cmake ~/Repos/iGenVar/ -DCMAKE_CXX_FLAGS="-Wall -Wextra -pedantic"` you will run into some warnings, which will also get fixed with this PR.

I added some tests and cleaned up the CLI tests in the process. Now the file is very difficult to review. So maybe just ignore it, since my new tests all go through and are not a big innovation.

### DIFF
--- a/include/variant_detection/method_enums.hpp
+++ b/include/variant_detection/method_enums.hpp
@@ -15,7 +15,7 @@ enum detection_methods
 
     //!\cond
     // ATTENTION: Must always be the last item; will be used to determine the number of ids.
-    SIZE //!< Determines the size of the enum.
+    // SIZE //!< Determines the size of the enum.
     //!\endcond
 };
 
--- a/test/cli/iGenVar_cli_test.cpp
+++ b/test/cli/iGenVar_cli_test.cpp
@@ -172,9 +172,11 @@ std::string expected_res_empty
 std::string expected_err_default_no_err
 {
     "Detect junctions in long reads...\n"
+    "The read depth method for long reads is not yet implemented.\n"
+    "The read depth method for long reads is not yet implemented.\n"
+    "The read depth method for long reads is not yet implemented.\n"
+    "The read depth method for long reads is not yet implemented.\n"
     "Start clustering...\n"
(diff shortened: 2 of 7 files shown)


test/cli/iGenVar_cli_test.cpp
```diff
@@ -282,7 +282,7 @@ TEST_F(iGenVar_cli_test, fail_negative_min_var_length)
                                          "--min_var_length -30");
     std::string expected_err
     {
-        "[Error] You gave a negative min_var_length parameter.\n"
+        "[Error] Value parse failed for --min_var_length: Argument -30 could not be parsed as type std::string.\n"
```
Irallia, line 285: Why is'nt this `size_t`? Is this a bug of the seqan3 parser?
```suggestion
        "[Error] Value parse failed for --min_var_length: Argument -30 could not be parsed as type size_t.\n"
```

Irallia: Resolved by https://github.com/seqan/seqan3/pull/2836

include/variant_detection/method_enums.hpp
```diff
@@ -15,7 +15,7 @@ enum detection_methods
 
     //!\cond
     // ATTENTION: Must always be the last item; will be used to determine the number of ids.
-    SIZE //!< Determines the size of the enum.
+    // SIZE //!< Determines the size of the enum.
```
Irallia, line 18: Since we don't need this at the moment I have just commented it out.

eseiler (approved):

PR #167, opened 2021-09-27
[INFRA] Add warning flags to library

eseiler:


src/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ add_library ("${PROJECT_NAME}_lib" STATIC modules/clustering/hierarchical_cluste
 target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC seqan3::seqan3)
 target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC fastcluster)
 target_include_directories ("${PROJECT_NAME}_lib" PUBLIC ../include)
+target_compile_options ("${PROJECT_NAME}_lib" PUBLIC "-pedantic" "-Wall" "-Wextra" "-Werror")
```
Irallia, line 20: I read that you should not use Werror, because why not get warnings as warnings, than errors: https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/
```suggestion
target_compile_options ("${PROJECT_NAME}_lib" PUBLIC "-pedantic" "-Wall" "-Wextra")
```

src/modules/clustering/hierarchical_clustering_method.cpp
```diff
@@ -223,6 +223,7 @@ std::vector<Cluster> hierarchical_clustering_method(std::vector<Junction> const
         // Add new clusters: junctions with the same label belong to one cluster
         for (auto & [lab, jun] : label_to_junctions )
         {
+            (void) lab;
```
Irallia, line 226: does this solve a warning?

Irallia (changes requested):
Thanks, I have one remark about Werror and a question.

src/CMakeLists.txt
eseiler: There are many reasons to use `-Werror` (the linux kernel tried to set this flag as default recently, too).

But it is quite annoying for developing; I removed it.

src/modules/clustering/hierarchical_clustering_method.cpp
eseiler: `lab` is otherwise unused, and you cannot use `[[maybe_unused]]` with structural bindings. Using `std::get<1>` would also be possible, but it doesn't make a difference, and the structural binding is easier to understand.

Irallia: ah thx for the explanation :)

Irallia (approved):
LGFM

joergi-w (approved):
Thank you, looks good!

PR #168, opened 2021-10-04
[FEATURE] Call tandem duplications from SA tag

### LINKED ISSUE #165: [FEATURE] Get Tandem Duplications from SA tag

Irallia:
Resolves #165


src/modules/sv_detection_methods/analyze_split_read_method.cpp
```diff
@@ -123,11 +134,67 @@ void analyze_aligned_segments(std::vector<AlignedSegment> const & aligned_segmen
...
+                    // current_segment ----------------------
+                    //                         ||||||||||||||||||||||
+                    // next_segment            ----------------------
+                    if (current.ref_name == next.ref_name &&
+                        current.get_reference_start() <= next.get_reference_start() &&
+                        next.get_reference_start() < current.get_reference_end())
+                    {
+                        tandem_dup_length_on_read += (current.get_reference_end() - next.get_reference_start() + 1);
+                        size_t single_dup_len{0};
+                        // If first pair of a tandem duplication ...
```
MitraDarja, line 150: Why the three dots?

src/variant_detection/variant_output.cpp
```diff
@@ -36,8 +36,26 @@ void find_and_output_variants(std::map<std::string, int32_t> & references_length
                         if (std::abs(sv_length) >= args.min_var_length &&
                             std::abs(sv_length) <= args.max_var_length)
                         {
+                            // Tandem Duplication
+                            if (
+                                // distance >= args.min_var_length &&
+                                // distance <= args.max_var_length &&
```
MitraDarja, line 42:
```suggestion
                                // distance <= args.max_var_length
```

test/api/detection_test.cpp
```diff
@@ -532,12 +532,168 @@ TEST(junction_detection, analyze_sa_tag)
...
+                    junctions_res[i].get_read_name()) << "Read names of junction " << i << " unequal";
+            EXPECT_TRUE(junctions_expected_res[i] == junctions_res[i])
+                        << "Junction " << i << " unequal\n"
+                        << "Mate 1 equal: " << (junctions_expected_res[i].get_mate1() == junctions_res[i].get_mate1())
+                        << "\nMate 2 equal: " << (junctions_expected_res[i].get_mate2() == junctions_res[i].get_mate2())
+                        << "\n";
+        }
+
+        // For debugging use:
+        // print_compare_junction_vectors(junctions_expected_res, junctions_res);
```
MitraDarja, line 586: Do you want to keep this? If so, maybe use a debugging option like if (debug) print...

test/api/detection_test.cpp
```diff
@@ -532,12 +532,168 @@ TEST(junction_detection, analyze_sa_tag)
...
+        std::vector<Junction> junctions_res{};
+
+        // pos  11  21           31                                                  41
+        //      |   |            |                                                   |
+        // chr1 -1->-----2------>-----2------>                                       -3->
+        //      ||||||||||||||||||||||||||||||\\\\\\\\\\\\ \\\\\\\\\\\\ \\\\\\\\\\\\ ||||
+        // read --->-DUP:TANDEM->-DUP:TANDEM->-DUP:TANDEM->-DUP:TANDEM->-DUP:TANDEM->--->
+        //          |            |            |            |            |            |
+        // read_pos 11           21           31           41           51           61
+
```
MitraDarja, line 600: Do you want all these prints in your api-test?

MitraDarja (changes requested):
LGTM, just a few minor things :)

test/api/detection_test.cpp
Irallia: I would like to keep it that way, because then I can print only the parts I really want to debug.

test/api/detection_test.cpp
Irallia: Thats just an explanation, how the example look like. It will not get printed.

PR #169, opened 2021-10-04
[TEST] Refactor api tests

Irallia:
Some refactoring for #168 .


test/api/api_test.hpp
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <gtest/gtest.h>
+
+#include <seqan3/core/debug_stream.hpp>
+
+#include "iGenVar.hpp"  // for global variable gVerbose
```
Irallia, line 7: Is there a way to set this global variable gVerbose to true here?

SGSSGene (changes requested):
There might be subtle problem with `gVerbose = true`. By default it is initialized to `false`. Depending on the unit test order, it might be set to true by a previous unittest.
So it is necessary to reset it to false.
I think this could be a great use case of the use of a scope guard (Idea taken from here https://stackoverflow.com/questions/28729545/abusing-c11-unique-ptr-to-execute-code-upon-leaving-the-scope):
By implementing a scope guard function:
```
template<class F> auto scope_guard(F&& f) {
    return std::unique_ptr<void, std::decay<F>::type>{(void*)1, std::forward<F>(f)};
}
```
You can now protect your code by writting:
```
TEST(...) {
    auto verboseGuard = scope_guard([]{gVerbose = false; });
    gVerbose = true;
    ....
}
```
Or you define even a help construct for that:
```
template<class F> auto scope_guard(F&& f) {
    return std::unique_ptr<void, std::decay<F>::type>{(void*)1, std::forward<F>(f)};
}
auto guard_verbose (bool verbose) {
    auto lastState = gVerbose;
    gVerbose = verbose;
    return scope_guard([=]{gVerbose = lastState; });
}
```
This would allow you to write:
```
TEST(...) {
    auto verboseGuard = verbose_guard(true); // will reset back to the original state, after leaving this scope
    ....
}
```

test/api/detection_test.cpp
```diff
@@ -426,178 +446,182 @@ TEST(junction_detection, analyze_sa_tag)
                        1000,                    // default partition_max_distance
                        0.5};                    // default hierarchical_clustering_cutoff
 
-    std::vector<Junction> junctions_res{};
+    { // Example 1
+        seqan3::debug_stream << "----------------------------------First Example:----------------------------------\n";
```
joergi-w, line 450: This is printed on each test execution – is this intended (even if `print_compare_junction_vectors` is just a comment)? The same below for the second example.

PR #170, opened 2021-10-14
[FIX] simplify passing file extensions to the validator

joergi-w:
Currently, we manually collect valid file extensions from the different file formats. I found a simpler way to achieve this. 

I found that we cannot read compressed sequence files because they are rejected by the validator, and SeqAn has already an issue for that: seqan/sharg-parser#77 
Maybe we can push this issue a bit more into attention, as also other apps (like Mars) suffer from that.


SGSSGene (approved):
I wasn't aware that `seqan3::input_file_validator<seqan3::sequence_file_input<>>{}` exists. This is very cool! :open_mouth:

PR #171, opened 2021-10-18
Calculate active regions based on Cigar information

### LINKED ISSUE #153: Determine the active regions in a genome
Use the raw activity profile:

Smooth the wave function, and determine the regions of high activity. 
Parameters: threshold of activeness, minimum and maximum length of regions

[Algorithmic details](https://gatk.broadinstitute.org/hc/en-us/articles/360036227652-ActiveRegion-determination-HaplotypeCaller-and-Mutect2-)

joergi-w:
I'm using the activity values from a previous PR to calculate the regions of high activity with a sliding window approach. The regions are returned in the format of intervals (i.e. pairs of genome positions). 

The first commit adds the new functionality, the second commit structures the already quite long function into smaller, testable functions. The third commit adds API tests in order to satisfy coverage tests. Therefore, it may be easier to **review them separately**.

Resolves #153.

### DIFF
--- a/src/variant_detection/snp_indel_detection.cpp
+++ b/src/variant_detection/snp_indel_detection.cpp
@@ -9,7 +9,7 @@
 
 void update_activity_for_record(std::vector<unsigned> & activity,
                                 std::vector<seqan3::cigar> const & cigar_sequence,
-                                uint64_t min_var_length,
+                                uint64_t const min_var_length,
                                 int32_t ref_pos)
 {
     // Track the current position within the read.
@@ -60,7 +60,7 @@ std::vector<std::pair<size_t, size_t>> active_regions(std::vector<unsigned> cons
 
     // Some constants for the sliding window approach.
     size_t const window_width = 5u; // another idea: (min_var_length + 1) / 2;
-    size_t const activity_threshold = 2;
+    size_t const activity_threshold = 2; // TODO (joergi-w 25.10.2021) depend on read coverage
 
     // Set up the result vector.
     std::vector<std::pair<size_t, size_t>> regions{};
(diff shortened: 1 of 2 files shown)


src/variant_detection/snp_indel_detection.cpp
```diff
@@ -47,48 +136,18 @@ void detect_snp_and_indel(std::filesystem::path const & reads_filename)
...
-            // Skip long variants.
-            if (length >= 30) // TODO (joergi-w 30.09.2021) Use the min_length parameter here.
-            {
-                read_pos += length;
-                ref_pos += length;
-                continue;
-            }
+        // Check if we have already allocated the length of the reference sequence.
+        if (activity_vec[ref_id].empty())
+            activity_vec[ref_id].resize(std::get<0>(header.ref_id_info[ref_id])); // length of reference sequence
```
joshuak94, line 141: Wouldn't this perform the resize multiple times, if you begin with several reads in a row which are just 151M? Because then `update_activity_for_record` doesn't perform any changes to the vector.

Or are redundant resizes not a problem?

joshuak94 (changes requested):

src/variant_detection/snp_indel_detection.cpp
joergi-w: No, I disagree: This resize is only executed once per reference id. For a file with 3 references, we initialize `activity_vec` as three empty vectors `[{}{}{}]` in line 120. The first time that a new ref_id appears in a record, we resize this one to the reference length (and implicitly fill zeroes). The `empty()` check returns `false` after resize!

More background why I implement it this way: I had a sam file of the human genome with 25 reference ids. However, it contained reads only of one chromosome, and we would unnecessarily allocate all the 25 vectors, although only in a single one there is relevant data.

PR #177, opened 2021-11-02
update seqan to master to include the file positioning information.

joshuak94:
The current branch of seqan3 in our submodule doesn't contain the necessary code for the bamintervaltree.


joshuak94:
@Irallia Any idea why updating seqan causes the argument_parser to try and parse all uint types as strings?

Irallia:
> @Irallia Any idea why updating seqan causes the argument_parser to try and parse all uint types as strings?

@eseiler I remember, that we had to do some changes regarding to differences between linux and mac. But now the argument_parser expects std::strings instead of integers. Have we done another change in seqan3?

joshuak94:
@Irallia I figured it out, it's because I'm updating the seqan submodule to the master branch, where the file seeking stuff is on. But the correct parsing for the argument parser is on the release branch. After the next merge everything should be fine.

joshuak94:
@Irallia Okay, now it should be good to merge.

Irallia (approved):

PR #178, opened 2021-11-03
Add BAMIntervalTree as a submodule.

joshuak94:


joshuak94:
This is still a work in progress!

joshuak94:
@Irallia I think the plan will be to make just the BAMIntervalTree submodule one pull request, and then actually use it in another pull request! So once this one gets merged, I will start working on the read depth method!

test/api/api_test.hpp
```diff
@@ -43,3 +45,15 @@ void print_compare_junction_vectors(std::vector<Junction> const & junctions_expe
                              << junctions_expected_res[i].get_tandem_dup_count() << " == " << junctions_res[i].get_tandem_dup_count() << "\n";
     }
 }
+
+// Helper function for comparing two bamit trees.
+void compare_bamit_trees(std::unique_ptr<bamit::IntervalNode> const & t1, std::unique_ptr<bamit::IntervalNode> const & t2)
```
joergi-w, line 50: Note that a failing EXPECT command does not stop the test execution and we always walk through each tree completely. In case you don't want to see a huge amount of error messages if two trees differ, this function could return a boolean whether they differ (and the walk can be pruned after the first difference) and then you check the function result with a single EXPECT statement. Another, more aggressive method is using `ASSERT_EQ` which aborts and fails the whole test immediately (I doubt that this is desired as you don't find further errors).

test/cli/iGenVar_cli_test.cpp
```diff
@@ -702,6 +706,7 @@ TEST_F(iGenVar_cli_test, dataset_single_end_mini_example)
     std::string output_err_str((std::istreambuf_iterator<char>(output_err_file)),
                                 std::istreambuf_iterator<char>());
     EXPECT_EQ(result.err, output_err_str);
+
```
joergi-w, line 709: I guess this change was unintentional

joergi-w (changes requested):
Thanks for this nice PR! I have a few suggestions:

test/api/api_test.hpp
```diff
@@ -43,3 +45,15 @@ void print_compare_junction_vectors(std::vector<Junction> const & junctions_expe
...
+{
+    if (!t1 && !t2) return;
+    if (!t1 || !t2) EXPECT_EQ(1, 0);
```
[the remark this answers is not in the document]
joshuak94: That's true, it's a bit confusing to write it this way. It's basically there so that if a node is NULL in one tree, but contains a valid interval in the other tree, it will fail the test. So I suppose the second option should work!

.gitmodules
```diff
@@ -2,3 +2,7 @@
 	path = lib/seqan3
 	url = https://github.com/seqan/seqan3.git
 	branch = master
+[submodule "lib/BAMIntervalTree"]
+	path = lib/BAMIntervalTree
+	url = git@github.com:joshuak94/BAMIntervalTree.git
```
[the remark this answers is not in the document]
joshuak94: Ah that makes sense haha.

test/cli/iGenVar_cli_test.cpp
joergi-w: You removed the wrong empty line (at the end of the file), but I don't mind much ;)

joergi-w (approved):
Thank you!

include/variant_detection/variant_detection.hpp
```diff
@@ -3,10 +3,15 @@
...
 #include <vector>
+#include <fstream>
 
 #include "iGenVar.hpp"              // for struct cmd_arguments
 #include "structures/junction.hpp"  // for class Junction
 
+#include "bamit/all.hpp"
+#include "cereal/types/memory.hpp"
+#include "cereal/types/vector.hpp"
+
```
Irallia, line 14: Please move all includes, which you don't need in the hpp to the cpp file, such that they are only loaded, when the cpp is needed.

test/cli/iGenVar_cli_test.cpp
```diff
@@ -726,4 +731,6 @@ TEST_F(iGenVar_cli_test, dataset_short_and_long_read_mini_example)
     std::string output_err_str((std::istreambuf_iterator<char>(output_err_file)),
                                 std::istreambuf_iterator<char>());
     EXPECT_EQ(result.err, output_err_str);
-}
+
+    std::filesystem::remove(short_reads_bamit_path);
+}
```
Irallia, line 736: please add the empty line

Irallia (changes requested):
This looks very good, thank you!
I will also look at the new repo :)
Here only 2 small changes, in general everything great!

Irallia (approved):
Thanks!

PR #186, opened 2021-11-27
[BENCHMARK] Repair script & update caller versions

Irallia:
I decided to add some results for a better comparison. We could split the Repo into a benchmark repo when we release iGenVar, but for now, I think its fine, to add everything here

The new Plot (I dont know why, but pbsv does get worse results when it trys to call DUPs):
[CallerComparisonPlot.pdf](https://github.com/seqan/iGenVar/files/7617441/CallerComparisonPlot.pdf)


joshuak94 (approved):

joergi-w (approved):
Thanks! I was unable to render the diff of the vcf files in my browser, so I didn't review them. Is it really necessary to commit these huge files (will users want to see them, although they're reproducible)? Note that every committed file will always be downloaded through `git clone` (even if removed/updated later on), which is especially important for large files.
Otherwise everything looks fine to me :)

Irallia:
> Thanks! I was unable to render the diff of the vcf files in my browser, so I didn't review them. Is it really necessary to commit these huge files (will users want to see them, although they're reproducible)? Note that every committed file will always be downloaded through `git clone` (even if removed/updated later on), which is especially important for large files. Otherwise everything looks fine to me :)

I would remove these files before a release, but for now they are very helpful, to  see directly any change, that appears with an update of one of the tools. As I wrote above, I would maybe split a benchmark repo out of iGenVar, and move some of these things there.

PR #190, opened 2022-02-04
 [INFRA] Remove gcc/g++ 7 and 8 support & Update seqan3 submodule

### LINKED ISSUE #187: [MISC] Update seqan3 to get thread problem fixed.
See https://github.com/seqan/seqan3/pull/2911 & https://github.com/seqan/seqan3/pull/2752

Irallia:
[MISC] Update seqan3 submodule to get thread problem fixed.
[MISC] Update filesystem includes
[INFRA] Remove gcc/g++ 7 and 8 support.

Resolves #187

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -47,16 +47,6 @@ jobs:
             build_type: Release
             cxx_flags: "-std=c++2a"
 
-          - name: "gcc8"
-            cxx: "g++-8"
-            cc: "gcc-8"
-            build_type: Release
-
-          - name: "gcc7"
-            cxx: "g++-7"
-            cc: "gcc-7"
-            build_type: Release
-
     steps:
       - name: Set repository name
         run: echo "REPOSITORY_NAME=$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')" >> $GITHUB_ENV
--- a/.github/workflows/ci_macos.yml
+++ b/.github/workflows/ci_macos.yml
@@ -40,16 +40,6 @@ jobs:
             build_type: Release
             cxx_flags: "-std=c++2a"
 
-          - name: "gcc8"
-            cxx: "g++-8"
-            cc: "gcc-8"
-            build_type: Release
-
-          - name: "gcc7"
-            cxx: "g++-7"
-            cc: "gcc-7"
-            build_type: Release
-
     steps:
       - name: Set repository name
         run: echo "REPOSITORY_NAME=$(echo '${{ github.repository }}' | awk -F '/' '{print $2}')" >> $GITHUB_ENV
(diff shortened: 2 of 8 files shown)


joergi-w (changes requested):
We are not affected by the "thread issue", since we set our own default number of threads:
```
src/iGenVar.cpp:    seqan3::contrib::bgzf_thread_count = args.threads;
```
Since I cannot find any improvement for iGenVar, I slightly disagree to drop gcc 7 and 8 compiler support for this. Anyway, updating SeqAn beyond the 3.1.0 release should be a careful consideration, as the update to a non-release version means a less stable interface and possibly new (yet undetected) errors. 
I guess we should discuss this step together in a meeting.

Irallia:
> We are not affected by the "thread issue", since we set our own default number of threads:
> 
> ```
> src/iGenVar.cpp:    seqan3::contrib::bgzf_thread_count = args.threads;
> ```

Yes, but this does not affect anything, because there was a bug in seqan3 which used all threads and ignores our given number. -> #187

> Since I cannot find any improvement for iGenVar, I slightly disagree to drop gcc 7 and 8 compiler support for this. Anyway, updating SeqAn beyond the 3.1.0 release should be a careful consideration, as the update to a non-release version means a less stable interface and possibly new (yet undetected) errors. I guess we should discuss this step together in a meeting.

SeqAn3 3.1.0 is the last SeqAn version which supports gcc 7 and 8, thus I decidet to drop the support aswell.

joergi-w:
> Yes, but this does not affect anything, because there was a bug in seqan3 which used all threads and ignores our given number. -> https://github.com/seqan/iGenVar/issues/187

Ah, I see. And SeqAn won't offer a 3.1.* release which fixes the bug (with keeping the compilers)? What's the general recommendation for all applications that use SeqAn3?

Irallia:
> > Yes, but this does not affect anything, because there was a bug in seqan3 which used all threads and ignores our given number. -> #187
> 
> Ah, I see. And SeqAn won't offer a 3.1.* release which fixes the bug (with keeping the compilers)? What's the general recommendation for all applications that use SeqAn3?

This seems to be a question for @seqan/core.

Irallia:
> > > Yes, but this does not affect anything, because there was a bug in seqan3 which used all threads and ignores our given number. -> #187
> > 
> > 
> > Ah, I see. And SeqAn won't offer a 3.1.* release which fixes the bug (with keeping the compilers)? What's the general recommendation for all applications that use SeqAn3?
> 
> This seems to be a question for @seqan/core.

I had a discussion with the core team. There is no regulation from the SeqAn side. But the recommondation is to drop gcc 7 and 8.
Both have no compiler support anymore. On top gcc12 comes up in April and than the compiler support of gcc9 ends aswell.
In SeqAn3 we have lots of workarounds for the old gcc versions, and as its a library it makes sense to support lots of compilers. But for an App we will offer precompiled binarys, thus we don't need a support of old compilers. And we want to avoid coding workarounds.
A last point to make: All distributions have gcc 9 at least, thus you assume that a user have access to it.

joshuak94 (approved):

joshuak94:
Sounds good! Dropping gcc7/8 seems okay to me. Like you said, gcc9 should be more or less ubiquitous by now.

joergi-w (approved):
Sounds convincing, thanks! 
Okay, then let's do it :)

PR #194, opened 2022-02-24
[FEATURE] Use b.i.o. lib for vcf parsing.

### LINKED ISSUE #193: [FEATURE] Use b.i.o. library for vcf parsing.
Given the new library https://github.com/h-2/b.i.o. we would like to use it instead of writing our own vcf parser.

Irallia:
Resolves #193

~Blocked by https://github.com/h-2/b.i.o./pull/29 and https://github.com/h-2/b.i.o./pull/31~


CMakeLists.txt
```diff
@@ -42,9 +42,14 @@ FetchContent_Populate(hclust)
 add_library ("fastcluster" STATIC ${hclust_SOURCE_DIR}/fastcluster.cpp)
 target_include_directories ("fastcluster" PUBLIC ${hclust_SOURCE_DIR})
 
+# Dependency: BAMIntervalTree.
 add_library ("bamit" INTERFACE)
 target_include_directories ("bamit" INTERFACE lib/BAMIntervalTree/include/)
 
+# Dependency: b.i.o.
+add_library ("bio" INTERFACE)
+target_include_directories ("bio" INTERFACE lib/b.i.o./include/)
```
h-2, line 51:
```suggestion
find_package (bio QUIET REQUIRED HINTS lib/bio/build_system)
```

src/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ add_library ("${PROJECT_NAME}_lib" STATIC modules/clustering/hierarchical_cluste
 target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC seqan3::seqan3)
 target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC fastcluster)
 target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC bamit)
+target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC bio)
```
h-2, line 21:
```suggestion
target_link_libraries ("${PROJECT_NAME}_lib" PUBLIC bio::bio)
```
But if it works like this, it probably doesn't matter.

src/variant_detection/variant_output.cpp
```diff
@@ -1,28 +1,104 @@
...
+    hdr.infos.push_back({ .id = "SVTYPE", .number = 1, .type = bio::var_io::value_type_id::string,
+                          .description = "Type of SV called."});
+    hdr.formats.push_back({ .id = "GT", .number = 1, .type = bio::var_io::value_type_id::string,
+                          .description = "Genotype"});
+
+    for (auto const & [id, length] : references_lengths)
+        hdr.contigs.push_back({ .id = id, .length = length});
+
+    // std::string filedate = "filedate=", transTime(),
+    hdr.other_lines = {"filedate=" + transTime(),
```
h-2, line 51: Does this work? I think you might need to add an s:
```suggestion
    hdr.other_lines = {"filedate="s + transTime(),
```
?

src/variant_detection/variant_output.cpp
```diff
@@ -1,28 +1,104 @@
...
-    header.add_meta_info("SVLEN", 1, "Integer", "Length of SV called.", "iGenVarCaller", "1.0");
-    header.add_meta_info("END", 1, "Integer", "End position of SV called.", "iGenVarCaller", "1.0");
-    header.print(references_lengths, args.vcf_sample_name, out_stream);
     for (size_t i = 0; i < clusters.size(); ++i)
     {
         size_t cluster_size = clusters[i].get_cluster_size();
+        // ignore low quality SVs
         if (cluster_size >= args.min_qual)
         {
+            bio::var_io::default_record<> rec;
```
h-2, line 87: If you write lots of records, it might be faster to create the record outside the loop and just overwrite the values inside.

h-2: Then you also only need to change the fields that you actually update.

h-2: [this only applies if you write the records directly, see below]

src/variant_detection/variant_output.cpp
```diff
@@ -94,24 +153,36 @@ void find_and_output_variants(std::map<std::string, int32_t> & references_length
...
 void find_and_output_variants(std::map<std::string, int32_t> & references_lengths,
                               std::vector<Cluster> const & clusters,
                               cmd_arguments const & args,
                               std::filesystem::path const & output_file_path)
 {
+    bio::var_io::header hdr;
+    write_header(references_lengths, args.vcf_sample_name, hdr);
+
+    std::vector<bio::var_io::default_record<>> records;
+    write_records(clusters, args, records);
```
h-2, line 165: I don't know if seed is important, but right now you are creating all records and adding them to a vector. And then you iterator over the vector to write each record. Why not write the records immediately when you create them?

h-2 (commented):
Looks good! Just some small suggestions

src/CMakeLists.txt
Irallia: This results in an error:
```bash
CMake Error at src/CMakeLists.txt:4 (add_library):
  Target "iGenVar_lib" links to target "bio::bio" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
...
```

src/variant_detection/variant_output.cpp
Irallia: Works both

CMakeLists.txt
Irallia: Thsi ends up in an error aswell:
```bash
[ 82%] Building CXX object src/CMakeFiles/iGenVar_lib.dir/variant_detection/variant_output.cpp.o
/Users/lydia/Repos/iGenVar/src/variant_detection/variant_output.cpp:8:10: fatal error: bio/var_io/writer.hpp: No such file or directory
    8 | #include <bio/var_io/writer.hpp>        // for bio::var_io::header
      |          ^~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

```
Do I have to do something else?

h-2: Hm, this is strange, and it should definitely work. Can you try doing the find_package before SeqAn's find_package? In fact, you shouldn't need SeqAn's find_package, because bio already finds SeqAn3.

src/CMakeLists.txt
h-2: Yeah, this is expected to only work, if the find_package worked.

.gitmodules
```diff
@@ -6,3 +6,7 @@
 	path = lib/BAMIntervalTree
 	url = https://github.com/joshuak94/BAMIntervalTree.git
 	branch = master
+[submodule "b.i.o."]
+	path = lib/b.i.o.
+	url = https://github.com/h-2/b.i.o..git
```
h-2, line 11:
```suggestion
	url = ../../h-2/b.i.o..git
```

I recently learned that you can use relative paths here, too. This respects choice to clone with ssh or https and has some other benefits. But it is also not really important 🙃

src/variant_detection/variant_output.cpp
```diff
@@ -1,117 +1,194 @@
...
-        find_and_output_variants(references_lengths, clusters, args, std::cout);
+        std::ostringstream stream{};
+        bio::var_io::writer writer{stream, bio::vcf{}};
+        writer.set_header(hdr);
+        for (size_t i = 0; i < clusters.size(); ++i)
+        {
+            // ignore low quality SVs
+            if (clusters[i].get_cluster_size() >= args.min_qual)
+            {
+                bio::var_io::default_record<> record{};
```
h-2, line 168: You might still be able to move this record out of the loop completely. But only if you always overwrite or clear the fields.

h-2 (approved):

CMakeLists.txt
```diff
@@ -24,8 +24,11 @@ string (ASCII 27 Esc)
 set (FontBold "${Esc}[1m")
 set (FontReset "${Esc}[m")
 
-# Dependency: SeqAn3.
-find_package (SeqAn3 QUIET REQUIRED HINTS lib/seqan3/build_system)
+# Dependency: b.i.o.
+find_package (bio QUIET REQUIRED HINTS lib/b.i.o./build_system)
+
+# # Dependency: SeqAn3. Will be found by b.i.o. aswell.
```
joergi-w, line 30:
```suggestion
# Dependency: SeqAn3 is found by b.i.o. as well.
```

src/variant_detection/variant_output.cpp
```diff
@@ -1,122 +1,159 @@
 #include "variant_detection/variant_output.hpp"
 
+#include <chrono>   // for std::chrono::system_clock
+#include <ctime>    // for std::localtime, std::time, std::time_t
 #include <iostream> // for std::cout
+#include <iomanip>  // for std::put_time
```
joergi-w, line 6: sort alphabetically?

test/cli/iGenVar_cli_test.cpp
```diff
@@ -209,8 +210,8 @@ TEST_F(iGenVar_cli_test, test_verbose_option)
         "Start clustering...\n"
     };
     EXPECT_EQ(result.exit_code, 0);
-    EXPECT_EQ(result.out.erase(32,19), expected_res_default);
     EXPECT_EQ(result.err, expected_err + expected_err_default_no_err_2);
+    EXPECT_EQ(result.out.erase(408,19), expected_res_default); // erase the filedate
```
joergi-w, line 214: The standard always uses a space after comma:
```suggestion
    EXPECT_EQ(result.out.erase(408, 19), expected_res_default); // erase the file date
```

joergi-w (approved):

PR #196, opened 2022-02-28
[INFRA] Update cmake version from 3.11.4 to 3.22.2

Irallia:


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.11)
+cmake_minimum_required (VERSION 3.22)
```
joergi-w, line 1: This is dangerous: Cmake v3.22 is younger than 3 months and almost everyone will run into build problems! FU has v3.18 on their servers, MPI has v3.16. Which functions of v3.22 are you actually using?

joergi-w (changes requested):

PR #199, opened 2022-03-29
[FIX] Interrupted BAMIT creation, which ends in an `cereal::Exception`.

Irallia:
If a run of iGenVar is aborted in the BAMIT creation, an incorrect index is stored, which cannot be read when iGenVar is called again:
```
terminate called after throwing an instance of 'cereal::Exception'
  what():  Failed to read 8 bytes from input stream! Read 0
Abort trap: 6
```
Fix: We write the index into a tmp file and save it with the correct filename after finishing.


joshuak94 (approved):
This is a nice solution to the issue!

SGSSGene (approved):

PR #200, opened 2022-03-30
[FIX] Remove the "chr" prefix of the ref name.

Irallia:
Chromosome names can differ between "1" and "chr1" in different input files. IGenVar can read more than one input file, thus we would distinguish same SVs from different input files if the chromosome naming is different. 
To avoid having to change the input file itself, we decided to make the change within iGenVar.
Thus we remove the "chr" prefix.


src/modules/sv_detection_methods/analyze_split_read_method.cpp
```diff
@@ -29,6 +29,12 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
         if (fields.size() == 6)
         {
             std::string ref_name = fields[0];
+            // As the chromosome names can differ between "1" and "chr1", we would distinguish same SVs from different
+            // input files if the chromosome naming is different. Thus we remove the "chr" prefix.
+            if (ref_name.rfind("chr", 0) == 0) { // pos=0 limits the search to the prefix
```
joergi-w, line 34: Interesting way of implementing this 😏 
It's fine, however I think that using `ref_name.starts_with()` would be more intuitive ([Link](https://en.cppreference.com/w/cpp/string/basic_string/starts_with)).

Irallia:
> Nice work! I have one major concern to think about: With SA tags like `1,101,+,6M94S,60,0;2,101,+,6S10M84S,60,0;1,107,+,16S10M74S,60,0;` and also the new vcf file, I find it much harder to read than before, because the `chr` helped at least to spot the sequence name. This leads me to the question: how about _adding_ the `chr` prefix where it is missing rather than removing it? A [regex_match](https://en.cppreference.com/w/cpp/regex/regex_match) could spot a single number in the reference name. What are possible disadvantages? Maybe we can discuss...

Yes, that is really a difficult question. Adding the "chr" shouldn't be any more difficult than removing it, and yes it would definitely be more readable. 
The only reason I decided to remove it was because the original truth set I use is without the "chr". But I have already created an edited truth set with "chr", so it wouldn't be such a big problem...

src/modules/sv_detection_methods/analyze_split_read_method.cpp
```diff
@@ -28,7 +28,9 @@ void retrieve_aligned_segments(std::string const & sa_string, std::vector<Aligne
         split_string(sa_tag, fields, ',');
         if (fields.size() == 6)
         {
-            std::string ref_name = fields[0];
+            // As the chromosome names can differ between "1" and "chr1", we would distinguish same SVs from different
+            // input files if the chromosome naming is different. Thus we add the "chr" prefix.
+            std::string const & ref_name = (fields[0].starts_with("chr") ? fields[0] : ("chr" + fields[0]));
```
joergi-w, line 33: This is not general enough I think. Imagine a reference name like "NC_000913.3 Escherichia coli str. K-12 substr. MG1655, complete genome" ([found here](https://www.ncbi.nlm.nih.gov/nuccore/NC_000913.3?report=fasta)), which doesn't start with `chr` and would turn into `chrNC_000913.3`. 
What we want to search is a single X or Y or at leat one digit, which is the regular expression `^([XY]|\d+)$`.

src/variant_detection/variant_detection.cpp
```diff
@@ -36,12 +36,18 @@ std::deque<std::string> read_header_information(auto & alignment_file,
...
 
     size_t i = 0;
-    for (std::string const & ref_id : ref_ids)
+    for (std::string & ref_id : ref_ids)
     {
+        // As the chromosome names can differ between "1" and "chr1", we would distinguish same SVs from different
+        // input files if the chromosome naming is different. Thus we add the "chr" prefix.
+        if (!ref_id.starts_with("chr")) { // pos=0 limits the search to the prefix
+            // Add "chr" prefix
+            ref_id = "chr" + ref_id;
```
joergi-w, line 49: the same here

src/variant_detection/variant_detection.cpp
```diff
@@ -36,12 +36,18 @@ std::deque<std::string> read_header_information(auto & alignment_file,
...
+    std::deque<std::string> ref_ids = alignment_file.header().ref_ids();
     std::vector<std::tuple<int32_t, std::string>> ref_id_info = alignment_file.header().ref_id_info;
 
     size_t i = 0;
-    for (std::string const & ref_id : ref_ids)
+    for (std::string & ref_id : ref_ids)
     {
+        // As the chromosome names can differ between "1" and "chr1", we would distinguish same SVs from different
+        // input files if the chromosome naming is different. Thus we add the "chr" prefix.
+        if (!ref_id.starts_with("chr")) { // pos=0 limits the search to the prefix
```
joergi-w, line 47: This comment is outdated

src/modules/sv_detection_methods/analyze_split_read_method.cpp
joergi-w: A temporary value like `"chr" + fields[0]` must be stored in a variable and not a reference.

joergi-w (changes requested):
Nice, the PR got indeed much smaller. I have a new remark (`starts_with` is not feasible any more) and a few smaller issues.

src/modules/sv_detection_methods/analyze_split_read_method.cpp
Irallia: I have a BAM file where you read exactly this in the header:
```
@SQ	SN:chr1	LN:248956422
@SQ	SN:chr2	LN:242193529
...
@SQ	SN:chr22	LN:50818468
@SQ	SN:chrX	LN:156040895
@SQ	SN:chrY	LN:57227415
@SQ	SN:chrM	LN:16569
@SQ	SN:chr1_KI270706v1_random	LN:175055
@SQ	SN:chr1_KI270707v1_random	LN:32032
...
@SQ	SN:chr22_KI270739v1_random	LN:73985
@SQ	SN:chrY_KI270740v1_random	LN:37240
@SQ	SN:chrUn_KI270302v1	LN:2274
@SQ	SN:chrUn_KI270304v1	LN:2165
@SQ	SN:chrUn_KI270303v1	LN:1942
...
@SQ	SN:chrUn_GL000218v1	LN:161147
@SQ	SN:chrEBV	LN:171823
```
So it is "chr" infront of everything.
(long_reads: HG002_PacBio_GRCh38.bam)
And I had a look into my other long read BAM files, and there was no mixed version (chr for 1 to Y and no chr for the others). Checking the short read files at the moment, but this will take a while.

Irallia: In the short read BAM I found
```
@SQ	SN:chrUn_JTFH01001996v1_decoy	LN:2009	AS:GRCh38_full_plus_hs38d1_analysis_set_minus_alts
@SQ	SN:chrUn_JTFH01001997v1_decoy	LN:2003	AS:GRCh38_full_plus_hs38d1_analysis_set_minus_alts
@SQ	SN:chrUn_JTFH01001998v1_decoy	LN:2001	AS:GRCh38_full_plus_hs38d1_analysis_set_minus_alts
```

joergi-w (approved):
Thank you!

Irallia:
> Will review in a bit! But reading through made me think of something, what's the check we have to make sure the BAM files are all from the same reference genome? So that we don't have a long read file aligned to hg19, and a short read file aligned to GRCh38 or something, which could cause some unexpected behavior and incorrect results?

Oh that's a good point, I've already encountered that exact problem. As far as I can see, some of my example BAM files (all HG002) are aligned to different references:
```
    MtSinai_PacBio:     data/reference/GCA_000001405.15_GRCh38_no_alt_analysis_set.fa
    PacBio_CCS_10kb:    data/reference/hs37d5.fa
    10X_Genomics:       data/reference/hg19.reordered.fa
    Illumina:           data/reference/GCA_000001405.15_GRCh38_no_alt_plus_hs38d1_analysis_set.fa
    Illumina_Mate_Pair: data/reference/GCA_000001405.15_GRCh38_no_alt_plus_hs38d1_analysis_set.fa
```
Since we do not use FASTA files so far, we do not check anything. But with vaquita LR I encountered the problem because I can specify only one reference and sometimes just 2 exist, when I have 2 input files.

But you're right, maybe we should exclude this case completely anyway. Do you have an Idea what can go wrong? What exactly happens if I have an "old" aligned BAM file to hg18 and a newer one to hg19?
And should we warn the user or forbid these things?

Irallia:
If we take 2 different references, can it happen that the position of a certain SV is different?
If this is the case, we should really prohibit different ones.

joshuak94:
> If we take 2 different references, can it happen that the position of a certain SV is different? If this is the case, we should really prohibit different ones.

Exactly! For example, the sequence "ACAGTCATCGT" might map to chr1:100 in one reference, but chr1:150 in another reference.

joshuak94 (approved):
Overall, the pull request looks fine. However, I'm not sure if we want to merge this, because of the discussion above about checking to make sure the reference genomes are the same anyways, and *also* because this would not work for other species, i.e. mouse genomes.

So I think just enforcing that all input BAMs are aligned to the same reference is the most robust solution.

Irallia:
> Overall, the pull request looks fine. However, I'm not sure if we want to merge this, because of the discussion above about checking to make sure the reference genomes are the same anyways, and _also_ because this would not work for other species, i.e. mouse genomes.
> 
> So I think just enforcing that all input BAMs are aligned to the same reference is the most robust solution.

That is a very good and important objection. I have now studied references in great detail and what problems can occur.
In general, you probably want to have the same reference for everything, but that's hard for us to check. Also, I think there are situations where that doesn't matter. For example, there are various fasta files for GRCh38 / hg38 (e.g. some with and some without mitochondria or other). These should be compatible with each other, because they all describe the same version. Therefore I would leave the PR as it is, but add a notification that reminds the user to consider this (if he enters two input files).

joshuak94 (approved):
Looks good!

PR #208, opened 2022-05-24
[FIX] Add duplicated sequences to output and fix vcf output

Irallia:
This PR adds the duplicated sequence to the output.
To discuss: What is the length of a tandem duplication. In this PR I decided to use the length of the duplicated sequence.
For example:
```
ref  NNNNN AAACCCGGG AAACCCGGG NNNNNN
read NNNNN AAACCCGGG AAACCCGGG AAACCCGGG AAACCCGGG AAACCCGGG NNNNNN
```
Than the duplication goes from (5, 23],
its duplicated sequence is AAACCCGGG
and the SVLEN is 9.


src/variant_detection/variant_output.cpp
```diff
@@ -66,58 +66,56 @@ void write_record(Cluster const & cluster,
...
+                    // Increment end by 1 because VCF is 1-based
+                    record.info().push_back({.id = "END", .value = mate1.position + 1});
+                    sv_length = insert_size;
+                    sv_type = "INS";
+                }
                 // The SVLEN is neither too short nor too long than specified by the user.
                 if (std::abs(sv_length) >= args.min_var_length &&
-                    std::abs(sv_length) <= args.max_var_length)
+                    std::abs(sv_length) <= args.max_var_length &&
+                    ((sv_type != "DUP" || sv_type != "INS") || insert_size <= args.max_tol_deleted_length) &&
```
joergi-w, line 113: I don't understand this condition, and `(sv_type != "DUP" || sv_type != "INS")` is always true.

Also note that our definition of SVLEN has impact here.

joergi-w (changes requested):
> To discuss: What is the length of a tandem duplication.

I had a quick look at the specification ([source](https://www.internationalgenome.org/wiki/Analysis/Variant%20Call%20Format/VCF%20%28Variant%20Call%20Format%29%20version%204.0/encoding-structural-variants/)) and I found that
    SVLEN = Difference in length between REF and ALT alleles
Therefore, I would change the length to 27 in your example. 

However, it feels a bit redundant to me that SVLEN is always END-POS, because it doesn't add information.

Irallia:
> > To discuss: What is the length of a tandem duplication.
> 
> I had a quick look at the specification ([source](https://www.internationalgenome.org/wiki/Analysis/Variant%20Call%20Format/VCF%20%28Variant%20Call%20Format%29%20version%204.0/encoding-structural-variants/)) and I found that SVLEN = Difference in length between REF and ALT alleles Therefore, I would change the length to 27 in your example.
> 
> However, it feels a bit redundant to me that SVLEN is always END-POS, because it doesn't add information.

Yes, but the SVLEN is not used everywhere, I found a discussion about INVs [here](https://github.com/samtools/hts-specs/issues/139). Maybe we just introduce a second INFO value beside the SVLEN. I will adjust the PR.

src/variant_detection/variant_output.cpp
Irallia: Ah damn my summary has ruined it 😅 thx for the hint

src/variant_detection/variant_output.cpp
```diff
@@ -66,58 +68,57 @@ void write_record(Cluster const & cluster,
...
     record.genotypes().push_back({ .id = "GT", .value = std::vector{"./."s}});
+    record.info() = {};
 
     if (mate1.orientation == mate2.orientation)
     {
         if (mate1.seq_name == mate2.seq_name)
         {
-            size_t insert_size = cluster.get_average_inserted_sequence_size();
             if (mate1.orientation == strand::forward)
             {
```
joergi-w, line 78: These could be merged into one condition

src/variant_detection/variant_output.cpp
```diff
@@ -66,58 +68,57 @@ void write_record(Cluster const & cluster,
...
-                        record.info().push_back({.id = "SVTYPE", .value = "INS"});
-                        found_SV = true;
-                    }
+                    record.info().push_back({.id = "SVLEN", .value = distance});
+                    record.info().push_back({.id = "iGenVar_SVLEN", .value = sv_length});
+                    record.info().push_back({.id = "SVTYPE", .value = sv_type});
+                    found_SV = true;
                 }
             }
         }
```
joergi-w, line 124: because they end in the same place.

joergi-w (approved):
Looks good, thank you!

src/variant_detection/variant_output.cpp
Irallia: You mean the two if cases?
I opened this PR because I actually want to call INVs. For calling INVs there will be an `else` for `if (mate1.orientation == mate2.orientation)`. So I would leave this as it is.

joergi-w: The three nested if cases actually... 
Of course, if you add else branches later, then it won't be possible.

PR #209, opened 2022-05-24
[BENCHMARKS] dataset comparison long read enhancement

Irallia:
# Illumina Mate Pair

all | DUP as INS 
:--:|:-----------:
Illumina_Mate_Pair results all | Illumina_Mate_Pair results DUP_as_INS all

Precision | Recall
:----------:|:-----:
Illumina_Mate_Pair results pr_all | Illumina_Mate_Pair results rec_all

# Illumina Paired End

all | DUP as INS 
:--:|:-----------:
Illumina_Paired_End results all | Illumina_Paired_End results DUP_as_INS all

Precision | Recall
:----------:|:-----:
Illumina_Paired_End results pr_all | Illumina_Paired_End results rec_all


joshuak94:
So it looks like using *just* long reads still works much better than using both short and long reads combined? Also what is the difference between mate pair and paired-end?

joshuak94 (approved):

Irallia:
> So it looks like using _just_ long reads still works much better than using both short and long reads combined? Also what is the difference between mate pair and paired-end?

Yes, I had a discussion with Knut today aswell. And yes it just looks like out short read calling is very bad. 
We need to find out whats wrong there.

PR #217, opened 2022-06-30
deBruijn graph for haplotype calling

joergi-w:
This PR implements a deBruijn graph that can be used for calling haplotypes from short reads with respect to a reference sequence. Inline documentation and tests are included. The program that uses this data structure is supposed to perform the following steps:
- initialize an empty graph with a reference sequence
- add reads
- check whether the graph is viable (has enough unique kmers and no loops)
- prune nodes and edges with limited support
- retrieve the haplotype sequences

For debugging and other purposes, the graph can be exported into a formatted file, so it can be visualized with the `dot` program. I'm using the LEMON graph library for implementing the graph.

### DIFF
--- a/src/structures/debruijn_graph.cpp
+++ b/src/structures/debruijn_graph.cpp
@@ -17,21 +17,23 @@ bool DeBruijnGraph::init_sequence(unsigned char kmer_len, seqan3::dna4_vector co
     viable = false;
     graph.clear();
     assert(len <= reference.size());
+
+    // Reserve memory for nodes and arcs, because we know the size a-priori.
     graph.reserveNode(static_cast<int>(reference.size() + 1 - len));
     graph.reserveArc(static_cast<int>(reference.size() - len));
 
-    lemon::ListDigraph::Node prev_node = lemon::INVALID; // the previously processed node
+    lemon::ListDigraph::Node prev_node = lemon::INVALID; // the previously processed node (initial state: none)
     for (size_t kmer : seqan3::views::kmer_hash(reference, seqan3::ungapped{len})) // iterate kmers
     {
         if (nodes(kmer) == lemon::INVALID) // the kmer is new
         {
             nodes.set(graph.addNode(), kmer); // add node
-            if (prev_node != lemon::INVALID)
-                arcs.set(graph.addArc(prev_node, nodes(kmer)), 0); // add arc
+            if (prev_node != lemon::INVALID) // if there exists a previous node
+                arcs.set(graph.addArc(prev_node, nodes(kmer)), 0); // add arc from previous to current node
 
-            prev_node = nodes(kmer); // update processed node
+            prev_node = nodes(kmer); // update the previously processed node
         }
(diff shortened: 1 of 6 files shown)


src/structures/debruijn_graph.cpp
```diff
@@ -0,0 +1,203 @@
...
+            for (lemon::ListDigraph::NodeIt node(graph); node != lemon::INVALID; ++node)
+            {
+                lemon::ListDigraph::InArcIt in_arc(graph, node);
+                lemon::ListDigraph::OutArcIt out_arc(graph, node);
+                bool in = in_arc == lemon::INVALID || (arcs[in_arc] < 2 && ++in_arc == lemon::INVALID);
+                bool out = out_arc == lemon::INVALID || (arcs[out_arc] < 2 && ++out_arc == lemon::INVALID);
+                if (in && out)
+                    ++num_unique;
+                else
+                    ++num_non_unique;
```
joshuak94, line 86: I have to admit, I don't quite understand how the unique nodes are calculated haha

src/structures/debruijn_graph.cpp
```diff
@@ -0,0 +1,203 @@
...
+            }
+            viable = num_non_unique < 4 * num_unique;
+        }
+    }
+    return *viable;
```
joshuak94, line 91: Is there a reason to use `*viable` over `viable.value()`?

src/structures/debruijn_graph.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+void DeBruijnGraph::prune(size_t threshold)
+{
+    // We need to store the nodes and arcs to be deleted, because the iterator gets confused otherwise.
+
+    { // Erase each arc with read support below threshold.
+        std::vector<lemon::ListDigraph::Arc> del;
```
joshuak94, line 99: Does this create copies of the arcs to be deleted?

joshuak94 (approved):
Nice! Looks cool :)

src/structures/debruijn_graph.cpp
joergi-w: Thanks, I have added two comments to clarify :)
Unique means that at most one read supports the ingoing and outgoing arcs of the node. So I need to check whether there is only ≤ 1 arc with cargo ≤ 1.

src/structures/debruijn_graph.cpp
joergi-w: Yes, a good reason: `viable.value()` performs a check whether `viable` is set, and `*viable` does not. The check is unnecessary, since I am sure here that a value is set.

src/structures/debruijn_graph.cpp
joergi-w: Yes, it is copied, but an arc is internally just a number. The copy is therefore very cheap.

Irallia (changes requested):
LGFM, maybe some more comments?

Irallia (approved):
Perfect, thank you!

PR #226, opened 2022-07-26
[BENCHMARK] Add SV Amounts.

Irallia:
See plots.
The short read plots need more data from other tools. This is in progress right now.


test/benchmark/results/AmountSVs.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
+echo iGenVar
+
+for VARIABLE in L1 L2 L3 S1 S2 S1L1 S1L2 S1L3 S2L1 S2L2 S2L3
```
joergi-w, line 5: `VARIABLE` is not a good name ;) 
It seems that it encodes MINQUAL, so maybe we can derive a better name. 
The following if-elif-...-else construct should be replaced with a case statement, which is much better readable.
```bash
case $VARIABLE in
    S1L1 | S2L2)
        MINQUAL=9
        ;;
    ...
    *)
        echo "Wrong variable: $VARIABLE!"
        ;;
esac
```

joergi-w (approved):

test/benchmark/results/AmountSVs.sh
Irallia: I changed VARIABLE to input_combination, as in the iGenVar only benchmarks.

PR #227, opened 2022-08-02
SNP and Indel calling

joergi-w:
The first commit extends the junction class by two members: (1) The "deleted sequence" contains the part of the reference sequence that is deleted in the reads. (2) The "quality" is a float value to be passed to the vcf file and represents how well a SNP or Indel is represented in the read data. I implemented a new constructor with the new members next to the existing one, because it is a less invasive change than editing all the existing constructor calls.

The second commit adds the `snp_indel` method to iGenVar. It is enabled by default and used if a genome file and short reads are provided. We use the de Bruijn graph to compute haplotype sequences from the reads in a high-activity region. The haplotypes are in turn aligned against the reference, and SNPs and indels are extracted from the alignment and printed with the vcf writer. Currently, they are appended to the end of the existing vcf file; however, we can think about a different way in the future.


joergi-w:
The failing CI is a set-up problem with gcc 11 on Mac, and apparently the responsible script is `/lib/seqan3/.github/workflows/scripts/install_via_brew.sh` in the SeqAn repository.

src/variant_detection/variant_output.cpp
```diff
@@ -176,5 +177,42 @@ void find_and_output_variants(std::map<std::string, int32_t> & references_length
...
+        rec.id() = seqan3::detail::to_string("igenvar_", ins ? "ins" : del ? "del" : "snp", "_", cnt++);
+        rec.ref() = seqan3::ranges::to<std::vector>(junction.get_deleted_sequence() | seqan3::views::to_char);
+        rec.alt() = {seqan3::detail::to_string(junction.get_inserted_sequence())};
+        rec.qual() = roundf(junction.get_quality() * 100) / 100.F; // round 2 digits
+        rec.filter() = {"PASS"};
+        rec.genotypes() = {};
+        rec.genotypes().push_back({ .id = "GT", .value = std::vector{"./."s}});
+        rec.info() = {};
+        writer.push_back(rec);
+    }
```
joergi-w, line 215: The bug is fixed (https://github.com/biocpp/biocpp-io/issues/53), so this code can be simplified.

PR #218, opened 2022-07-08
[MISC] Update SeqAn3 and use sharg parser

Irallia:
This PR updates SeqAn3 to use the sharg parser.
As the b.i.o. lib is using an older version of SeqAn3 I had to change the dependency stuff in the CmakeLists.

Wait for https://github.com/seqan/sharg-parser/pull/111

PR #235, opened 2022-09-29
[INFRA] Update cmake version from 3.11.4 to 3.16.0

Irallia:
For https://github.com/seqan/iGenVar/pull/218 we need to update the cmake (>=3.16) version.

### DIFF
--- a/.github/workflows/ci_linux.yml
+++ b/.github/workflows/ci_linux.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 env:
-  CMAKE_VERSION: 3.11.4
+  CMAKE_VERSION: 3.16.0
   SEQAN3_NO_VERSION_CHECK: 1
   TZ: Europe/Berlin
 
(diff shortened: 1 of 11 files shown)


smehringer (approved):

.github/workflows/ci_linux.yml
```diff
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 env:
-  CMAKE_VERSION: 3.11.4
+  CMAKE_VERSION: 3.16.0
```
eseiler, line 11: This should be 3.16.9 (latest patch level). Googletest will require at least 3.16.3 when updated

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.11)
+cmake_minimum_required (VERSION 3.16)
```
eseiler, line 2: Strictly speaking, you don't need to modify the version here. You don't use any features that require a higher version. SeqAn will already check for 3.16

eseiler (changes requested):

Irallia:
I have accidentally merged, I will do the change to 3.16.9 in https://github.com/seqan/iGenVar/pull/218

